PR #3, opened 2015-09-30
Finished Assessment

dhuddell:
Update code within assessment.js to match requirements.
Update answers in README.md with my answers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,11 @@ You should save your answers in this README.md file, except for your answer to q
 
 Please list the JavaScript primitives and give one example of each.
 
-<!-- Replace this comment with your answer -->
+Number (1, -0)
+Boolean (true, false)
+String ('yo, yoyo')
+Null (null)
+Undefined (Undefined)
 
 ## Question 2
 
@@ -28,7 +32,8 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
-<!-- Replace this comment with your answer -->
+7
+c =(gets, defined as) c +(sum) d
 
 ## Question 3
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -28,7 +32,8 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
-<!-- Replace this comment with your answer -->
+7
+c =(gets, defined as) c +(sum) d
```
GA-MEB, line 36: That explanation could be more clear/explicit, but your reasoning is right.

README.md
```diff
@@ -53,19 +60,15 @@ weather === "sunny";
 
 What are the values of these expressions?  Explain your answers.
 
-<!-- Replace this comment with your answer -->
+First, weather is declared as a variable = undefined.
+Next, for weather = "sunny" (weather is defined with string "sunny").
+Then the final expression asks "is weather equal to sunny?" and
+it evalutes the boolean 'true'.
```
GA-MEB, line 66: Yep. Just keep in mind that the first two lines are expressions too, evaluating to `undefined` and `"sunny"`, respectively.

assessment.js
```diff
@@ -1,8 +1,8 @@
 //We'll learn about require later in the course
 var ask = require('./ask.js');
 
-var answer = 'not empty';
+var answer = 0;
 
-while (answer !== '' && answer !== 'SeCrEt') {
-  answer = ask("Guess my secret? ");
+while (answer > 10 || answer < 1) {
+  answer = ask("Guess within the range of my secret number. ");
```
GA-MEB, line 7: I think the question actually meant that you should have a secret number between 1 and 10, and that people should try to guess it. But this solution is otherwise reasonable.

GA-MEB:
5 of 5 

Nice work.

assessment.js
GA-MEB, line 7: One other comment here. `ask` will give you a string, not a number, so you'll need to perform that conversion in order for this to work properly.

PR #5, opened 2015-09-30
Assessment completed

julie1013:


README.md
```diff
@@ -53,7 +58,9 @@ weather === "sunny";
 
 What are the values of these expressions?  Explain your answers.
 
-<!-- Replace this comment with your answer -->
+weather = "sunny"; means that weather is assigned a value of "sunny".
+
+weather === "sunny" means that the variable weather evaluates to "sunny"
```
GA-MEB, line 63: Strictly, what this question was asking was the value that each expression evaluated to. 
`weather = "sunny"` not only assigns `weather` the value `"sunny"` - the expression itself also evaluates to `"sunny"`. 
`weather === "sunny"` evaluates to a boolean (in this case, `true`)

assessment.js
```diff
@@ -1,8 +1,7 @@
 //We'll learn about require later in the course
 var ask = require('./ask.js');
+var i = 1;
 
-var answer = 'not empty';
-
-while (answer !== '' && answer !== 'SeCrEt') {
-  answer = ask("Guess my secret? ");
+while (i >= 1 && i <= 10) {
+  console.log(i); i++;
```
GA-MEB, line 6: Not quite what was asked, though admittedly the question could have been worded more clearly. We were looking more for something along the lines of 

``` javascript
var i = -1;
while (i !== 7) {
  answer = ask("Guess a number between 1 and 10.");
}
```

GA-MEB:
4 of 5

Nice work!

PR #6, opened 2015-09-30
JS Basics Assessment Response

baz1389:


README.md
```diff
@@ -53,7 +53,9 @@ weather === "sunny";
 
 What are the values of these expressions?  Explain your answers.
 
-<!-- Replace this comment with your answer -->
+weather = "sunny"; will evaluate to the string 'sunny' since only one '=' is being used which will assign the variable weather a value of "sunny".
```
GA-MEB, line 56: That's right.

README.md
```diff
@@ -53,7 +53,9 @@ weather === "sunny";
...
+
+weather === "sunny"; will return undefined. Since we are using strict equality both sides have to be identical in data type and value.
```
GA-MEB, line 58: This is incorrect, though. We just assigned `weather` the value `"sunny"` in the previous line, so the two are equal; this expression should evaluate to `true`.

assessment.js
```diff
@@ -1,8 +1,6 @@
...
-var ask = require('./ask.js');
-
-var answer = 'not empty';
-
-while (answer !== '' && answer !== 'SeCrEt') {
-  answer = ask("Guess my secret? ");
+for (var i = 1; i < 11; i++){
+  if (i === 4){
+    console.log("Four!")
+  }
```
GA-MEB, line 5: Not quite what we were looking for. The program should allow users to guess until they get the number, e.g.

``` javascript
var answer = ''";
while (answer !== '4') {
  answer = ask("Guess a number between 1 and 10");
}
```

GA-MEB:
3.5 of 5

Nice work.

PR #7, opened 2015-09-30
morning JS assessment

faetea:
referred to my notes

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ You should save your answers in this README.md file, except for your answer to q
 
 Please list the JavaScript primitives and give one example of each.
 
-<!-- Replace this comment with your answer -->
+Number = 1, boolean = true, null, undefined, string = 'pickles'
 
 ## Question 2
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -28,7 +28,7 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
-<!-- Replace this comment with your answer -->
+c = 7, operators are +=
```
GA-MEB, line 31: Just remember that `=` isn't 'equals', like in math, it's 'assign/let', 
i.e. 
Let c be the sum of c and d.

README.md
```diff
@@ -53,7 +53,7 @@ weather === "sunny";
 
 What are the values of these expressions?  Explain your answers.
 
-<!-- Replace this comment with your answer -->
+unsure, I forget the difference between = and ===
```
GA-MEB, line 56: See above re: `=`. 
`===` is the strict equality comparison. It takes two arguments and evaluates to either `true` or `false` based on whether or not the two arguments are exactly equal.

As a side note, though, `===` cannot be used to compare arrays or objects.

GA-MEB:
4 of 5

Nice work!

PR #8, opened 2015-09-30
JS Basics Assessment

LaurenGrant:


README.md
```diff
@@ -53,7 +58,7 @@ weather === "sunny";
 
 What are the values of these expressions?  Explain your answers.
 
-<!-- Replace this comment with your answer -->
+I am unclear of what values are in this question. I know that the 1 = operator means that the varable weather is the string "sunny". the 3 ==== means that the variable weather is strictly eqaul to the string "sunny".
```
GA-MEB, line 61: In addition to corresponding to some sort of behavior, each line of code is an expression (e.g. 1+1), meaning that it can be evaluated to a single value.

The first expression, `var weather` yields the value `undefined`, as all `var` declarations are.
The second expression, `weather = "sunny"`, is an assignment; assuming that there's no `var`, an assignment operation evaluates to the value that's being assigned - in this case, "sunny".
The third expression, `weather === "sunny"`, is a comparison operation; since weather was assigned the value "sunny" in the previous line, this line should evaluate to `true`.

assessment.js
```diff
@@ -1,8 +1,5 @@
...
-var ask = require('./ask.js');
-
-var answer = 'not empty';
-
-while (answer !== '' && answer !== 'SeCrEt') {
-  answer = ask("Guess my secret? ");
+for (var i = 0; i < 10; i++) {
+  if (i===7) {
+    console.log("seven");
+  }
```
GA-MEB, line 4: This won't let people guess the number. Here's what we had in mind:

``` javascript
var answer = 'not empty';
while (answer !== '' && answer !== 3) { // 3 is the secret number
  answer = ask("Guess my secret! Pick a number between 1 and 10.");
}
```

GA-MEB:
3 of 5

Nice work!

PR #24, opened 2015-09-30
assesment.js

drewster317:


README.md
```diff
@@ -41,7 +45,7 @@ y = 10;
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- Replace this comment with your answer -->
+The Value of x and y are 10. The variables do not change but that values do.
```
GA-MEB, line 48: Not quite. `x = y` does not form any sort of persistent relationship between `x` and `y`; instead, `y` is evaluated first, and the value of `y` gets assigned to `x`. So although `y` becomes 10 in the last line, `x` doesn't change!

README.md
```diff
@@ -53,19 +57,21 @@ weather === "sunny";
 
 What are the values of these expressions?  Explain your answers.
 
-<!-- Replace this comment with your answer -->
+Value is True. It is a Boolean.
```
GA-MEB, line 60: Yep. What about the first two, though?

README.md
```diff
@@ -53,19 +57,21 @@ weather === "sunny";
...
 
 ## Question 5
 
 ```js
-
+var count = 0;
```
GA-MEB, line 67: It doesn't look like you're using `count` for anything - is that right?

README.md
```diff
@@ -53,19 +57,21 @@ weather === "sunny";
...
 var answer = 'not empty';
 
-while (answer !== '' && answer !== 'SeCrEt') {
-  answer = ask("Guess my secret? ");
+while (answer !== '7' ) {
+  answer = ask("Guess a number between 1 and 10?");
+
 }
+console.log ("You correctly answered it!")
```
GA-MEB, line 74: Otherwise, this looks good!

GA-MEB:
4 of 5 (-0.5 on Q3, Q4)

Nice work, Drew!

PR #29, opened 2015-09-30
Assessment

HoldenMills:


README.md
```diff
@@ -16,7 +16,11 @@ You should save your answers in this README.md file, except for your answer to q
 
 Please list the JavaScript primitives and give one example of each.
 
-<!-- Replace this comment with your answer -->
+Boolean - Truthy
+Strings - "This is a String"
+Numbers - 45
+Undefined - NaN //*note* This is wrong, but I only figured that out while looking something else up.//
```
GA-MEB, line 22: That's right - bizarrely, NaN itself _is actually a Number_.

README.md
```diff
@@ -41,7 +45,11 @@ y = 10;
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- Replace this comment with your answer -->
+x = 4
+y = 3
+x = 4 : y= 4
```
GA-MEB, line 50: Not quite. `x = y` takes y's value and copies it into x. As of this line, y is equal to 3, so that means that x will be set to 3.

README.md
```diff
@@ -53,7 +61,7 @@ weather === "sunny";
 
 What are the values of these expressions?  Explain your answers.
 
-<!-- Replace this comment with your answer -->
+weather === "sunny" is a strict assignment.  weather = "sunny" is more loose.
```
GA-MEB, line 64: No, sorry. 
`weather = "sunny"` is the assignment - we're assigning `weather` the value "sunny".
`weather === "sunny"` is a strict _comparison_ - it will compare the two values on either side and evaluate to either `true` or `false` depending on whether or not the two values are **exactly** equal (not 'sorta' equal, which is what the loose equality test, `==`, will check for).

README.md
```diff
@@ -63,7 +71,7 @@ var ask = require('./ask.js');
 
 var answer = 'not empty';
 
-while (answer !== '' && answer !== 'SeCrEt') {
+while (answer !== (1<10)) {
```
GA-MEB, line 74: This line doesn't do what you think it does. 
`1 < 10` evaluates to `true`, so what you're really writing is
`while (answer !== true)`
which I don't think was your intention.

assessment.js
```diff
@@ -1,8 +1,8 @@
 //We'll learn about require later in the course
 var ask = require('./ask.js');
 
-var answer = 'not empty';
+var answer = x;
 
-while (answer !== '' && answer !== 'SeCrEt') {
+while (answer == (x>1 && x<10) {
   answer = ask("Guess my secret? ");
 }
```
GA-MEB, line 8: This answer also has issues. The expression `(x > 1 && x < 10)` will evaluate to true when x is between 1 and 10, but you never actually change the value of x; as a result, the loop runs forever. Since you're getting new `answer` values each time you prompt the user, why not just write  `while (answer > 1 && answer < 10)`.

You're also doing a loose equality test (`==`), which you should avoid. And you actually never set the value of x before you used it in line 4 - this would give you an error if you were to run it.

GA-MEB:
2.5 of 5

Not bad, Holden, but I think you can do better with a little practice. Let me know if you want to try resubmitting this one - I'd be happy to give you some more feedback.

PR #33, opened 2015-10-05
Response _ KMcMahon

resource11:
Looks like I never created a pull request for this last week?

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,16 @@ You should save your answers in this README.md file, except for your answer to q
 
 Please list the JavaScript primitives and give one example of each.
 
-<!-- Replace this comment with your answer -->
+The five JavaScript primitives are number, string, boolean, null, and undefined
+
+Examples:
+
+2 is a number
+"mary" is a string
+true is a boolean
+null is... null!
+undefined is undefined
+
 
 ## Question 2
 
@@ -28,7 +37,12 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
-<!-- Replace this comment with your answer -->
+The value of c after the code executes is 7.
+
+In the last line, the operator "=" is an assignment operator. In this case, the "=" operator assigns to c the result of what the expression to the right of the "=" operator evaluates to.
+
+the "+" operator (I believe) is the addition operator rather than the numeric conversion (unary?) operator because we are evaluating two variables that have been assigned number values. If the variable c was initialized with a string value, I think that '+' would be considered a numeric conversion operator. I'm a bit shaky on this concept---numeric conversions is a new concept!
+
 
 ## Question 3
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -28,7 +37,12 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
-<!-- Replace this comment with your answer -->
+The value of c after the code executes is 7.
+
+In the last line, the operator "=" is an assignment operator. In this case, the "=" operator assigns to c the result of what the expression to the right of the "=" operator evaluates to.
+
+the "+" operator (I believe) is the addition operator rather than the numeric conversion (unary?) operator because we are evaluating two variables that have been assigned number values. If the variable c was initialized with a string value, I think that '+' would be considered a numeric conversion operator. I'm a bit shaky on this concept---numeric conversions is a new concept!
+
```
GA-MEB, line 45: Very good! Yes, the behavior of `+` is  dependent on the kinds of inputs it's given. 
3 + 3 => 6   (add)
'3' + '3' => '33' (concatenate)
'3' + 3 => '33' (convert and concatenate)
3 + '3' => '33' (convert and concatenate)

For some bizarre, extreme examples, check this out:
https://www.destroyallsoftware.com/talks/wat

README.md
```diff
@@ -41,7 +55,9 @@ y = 10;
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- Replace this comment with your answer -->
+OK. So the variable x is assigned 4 in the first line and the variable y is assigned 3 in the second line. When these two lines run in the node REPL environment, they both return a value of undefined, but if I console.log(x) and console.log(y) they console prints out the values assigned to x and y, which are 3 and 4 respectively.
+In the third line, x is assigned the value of y. Therefore, x results to 3.
+In the fourth line, y is assigned the value of 10, to x results to 10.
```
GA-MEB, line 60: Not quite. `x = y` doesn't establish any permanent relationship between x and y. All it does is (a) evaluate y, which gives a result of 3, and then subsequently, (b) take the value of 3 and assign it to x. If we then change the value of y, the value of x is unaffected.

assessment.js
```diff
@@ -1,8 +1,8 @@
 //We'll learn about require later in the course
 var ask = require('./ask.js');
 
-var answer = 'not empty';
+var answer = 0;
 
-while (answer !== '' && answer !== 'SeCrEt') {
+while (answer <= 7 && answer >= 9) {
```
GA-MEB, line 6: This condition will actually never be true. 7 satisfies the first condition but not the second; 9 satisfies the second condition but not the first; 8 satisfies neither condition. What you probably meant was to use an **or** operator (`||`) rather than an **and** operator (`&&`); in that case, the program would keep guessing until someone guessed 8. 

Also, one other thing to consider is that the result value from `ask` will be a String, so you'll need to either convert that to a Number or handle it as a String.

GA-MEB:
4 of 5 (-0.5 on Q3, Q5)

Nice work - let me know if you have questions about any of my feedback.

assessment.js
resource11, line 6: Uhhh... yeah. I really didn't completely understand clearly what was needed here. Hence the supreme lack of logic in this case. I guess I misread the first line of the question completely (I thought we were evaluating for numbers only, rather than strings).

README.md
resource11, line 45: Just getting caught up... That URL. Hilariousl

README.md
resource11, line 60: Understood. I wasn't clear that you wanted me to detail that there _is_ no permanent relationship between x and y going forward. Suggestion for next cohort: add another sentence in your bevy of questions asking what the value of x is after y is assigned the value of 10. That may make a student think a bit deeper on this. Does that makes sense at all?

assessment.js
resource11, line 6: So... would this work better?

 var answer = 0;
// var theNumber = 5;

 while (answer < 5 && answer >= 6) {
   answer = ask("Guess which number I'm thinking of between 1 and 10?");
 }

console.log("Yes! I was thinking of 5. Good work!");

PR #43, opened 2016-01-21
js-basics-diagnostic

apalmer0:


RealWeeks:
4/5.  What if it's not a number?  Good start though.

PR #47, opened 2016-01-21
Complete diganostics for js basics

dumplingpuff:


RealWeeks:
4/5.  Close, I see what you tried to do, but you didn't break the while so I'm caught in the loop.

PR #49, opened 2016-01-21
A. Jordon Diagnostic response

ajordon:


RealWeeks:
3/5.
Forgot the primitive string.  Also I see what you tried to do but there is no else in a while statement.  Next time use node in your console to check you answer, it blows up.  But nice shot, a few small changes and it would would great.

PR #50, opened 2016-01-21
Diagnostic Complete

edesilets:


RealWeeks:
4/5.  Good code, I know we didn't ask you but for a challenge if you what to go back and work on it try making it so it doesn't break out of the while statement if i enter a blank string.

Also:

ES5 has 5 they are as folows number (2), string ('I love dogs'), fourth thing , fifth thing.
ES6 added symbols.

really? 4th thing and 5th thing.  This link may help you next diagnostic: http://bfy.tw/3qNE

PR #51, opened 2016-01-21
js basic diagnostic 1/21

msyao:


RealWeeks:
4/5. Good job , but a few things.
while (answer !== '>=1','<=10') should be written: 

while(answer !== '' && answer <=10) - see what i did there? you need the variable and && operator.

PR #54, opened 2016-01-21
Response

rchezrutt88:


RealWeeks:
4/5 doesn't break the while.  Also it's recommended that you try to factor out using a continue statement.

PR #55, opened 2016-01-21
James Friedman JS-basics diagnostic response

jpfriedman18:
For some reason my code for #5 isn't working as expected (though it does run). It won't break when the correct number is given. Please let me know what I did wrong.


RealWeeks:
4/5. Close, but never break there while statement.  Given the discussion on slack if you were to do it again I would assume that it would be a cake walk.

PR #56, opened 2016-01-21
js-basics-diagnostics answers

kpie89:


RealWeeks:
Close.  Looks like you got caught here:

if (answer === (>1 || <10)) 

we went over this in slack but in case

if (answer === '' || answer <= 10)

you need the variable on both sides of the comparison operators.

PR #57, opened 2016-01-21
Response

jdamiron:


RealWeeks:
4/5.  

You blow up here: (answer !=== >=1 && <=10) 

just need to do something like: (answer !== >=1 && answer <=10) 

notice the !== and answer on both sides of the comparison.

PR #59, opened 2016-01-21
Diagnostic JS Basic 01/21/2016

kenvng:
Please let me know what I did wrong to the last question which is the programmingcoding one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,7 @@ Assume all code executes in strict mode.
 ## Question 1
 
 Please list the JavaScript primitives and give one example of each.
-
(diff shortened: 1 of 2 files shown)


RealWeeks:
4/5.  Your code blows up,  try running it in node first and if you have any questions or issues include a comment to i know what to look at.  We discussed a lot of this in slack already let me know if you want any help on this later or want to resubmit for me to look at.

I think you misunderstood question 4 because you didn't take into account that the variable for weather was stored in the previous line, but I can tell you understand the concept because you explained the strict equality check.

PR #60, opened 2016-01-21
did the test

Knguyen21:


RealWeeks:
4/5.  I have to go by what you submitted but your code looks great and we talked about it recently.  You went above the call here and that's awesome.  I think you're final solution is the best in the class.  You'll be a great asset as the class progresses.

PR #71, opened 2016-03-23
Complete js-basics-diagnostic

MichaelB167:


bin/diagnostic.js
```diff
@@ -1 +1,15 @@
 'use strict';
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+  if (answer (!>=1 && !<=10)) {
```
RDegnen, line 8: the correct way to write this would be `(answer < 1 || answer > 10)`

RDegnen:
4 out of 5 not bad

PR #72, opened 2016-03-23
Answer to Diagnostic

rdelvallej32:
Both ReadMe and bin file are updated with the new answers and code.

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,13 @@
 'use strict';
+
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '7') {
+  answer = ask("Guess a number between 1 and 10 ");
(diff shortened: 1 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,13 @@
...
+
+let answer = 'not empty';
+
+while (answer !== '7') {
+  answer = ask("Guess a number between 1 and 10 ");
+  if (answer === '7') {
+   console.log("You guessed it! You should play the Powerball");
+   break;
+  }
+}
```
RDegnen, line 13: You forgot to ask if the number is between 1 and 10. You can check by writing either `( answer < 1 || answer > 10)` to check if the answer falls outside that range OR `(answer >= 1 && answer <= 10)` to see if answer falls inside that range.

RDegnen:
4 out of 5 not bad

PR #75, opened 2016-03-23
js- Diagnostic

pswope:
Added answers

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,15 @@
 'use strict';
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '') {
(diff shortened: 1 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,15 @@
...
+
+let answer = 'not empty';
+
+while (answer !== '') {
+  answer = ask("Guess my secret number between 1 and 10? ");
+  if (answer === '8') {
+   console.log("You guessed it!");
+   break;
+  }
+}
```
RDegnen, line 12: You forgot to ask if the number is between 1 and 10. You can check by writing either `( answer < 1 || answer > 10)` to check if the answer falls outside that range OR `(answer >= 1 && answer <= 10)` to see if answer falls inside that range.

RDegnen:
4 out of 5 not bad

PR #80, opened 2016-03-23
diagnostic-submission js-basics

BenGitsCode:
I made that more complicated than it was but I
think I'm just not used to using node to test atom files yet.

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,18 @@
 'use strict';
+
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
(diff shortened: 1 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,18 @@
...
+  if (answer === 8) {
+   console.log("You guessed it!");
+   break;
+  }
+if (answer >= 1 && answer <= 10) {
+  console.log('Try Again ');
+} else if (answer !== 8) {
+  console.log('Your guess needs to be between 1 and 10 doofus');
+}
+}
```
RDegnen, line 18: this is ok, but using two if statements is not what you want here, if should all be one if statement.

RDegnen:
4 out of 5 not bad

PR #81, opened 2016-03-23
adds answers

V4Vince:


README.md
```diff
@@ -18,6 +18,13 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+Boolean = true or false
+Null = the value is nothing
+Undefined = No value
+number = all integers
+string = all letters including the $
```
RDegnen, line 25: a better way to state this would be anyhing between quotes so `" "` is still a string and so is `"12"`

RDegnen:
5 out of 5

PR #82, opened 2016-03-23
Response

louarnos:


bin/diagnostic.js
```diff
@@ -1 +1,19 @@
...
+   console.log("Please Guess a number between 1 and 10!");
+ }
+ else if (answer === myNum) {
+   console.log("You guess correct!")
+   break;
+ }
+ else {
+   console.log("Not my number! Guess again");
+ }
+  }
```
RDegnen, line 19: this wasn't working because myNum was set to a number and not a string

RDegnen:
4 out of 5 not bad

PR #83, opened 2016-03-23
Response to js-basics questions

ghost:


README.md
```diff
@@ -64,10 +75,13 @@ const ask = require('../lib/ask.js');
 let answer = 'not empty';
 
 while (answer !== '') {
-  answer = ask("Guess my secret? ");
-  if (answer === 'SeCrEt') {
+  answer = ask("Guess my secret number? ");
+  if(answer >= 1 && answer <=10) {
    console.log("You guessed it!");
    break;
+  if(answer <= 1 && answer >= 10) {
```
RDegnen, line 82: this won't work because answer can't be less than or equal to 1 AND  greater than or equal to 10. you would write this like so `( answer < 1 || answer > 10 )` which is is answer less than 1 OR greater than 10?

RDegnen:
4 out of 5

PR #93, opened 2016-05-18
js-basics-response (README.md & diagnostics.js)

RoobScoobs:
Includes all 5 questions & answers

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,15 @@
 'use strict';
+
+const ask = require('../lib/ask.js');
+
+let answer = 7;
+
+while (answer !== 7) {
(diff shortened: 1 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,15 @@
...
+
+while (answer !== 7) {
+  answer = ask("Guess my secret? ");
+  if (answer === 7) {
+   console.log("You guessed it!");
+   break;
+ } else if (answer !== 1 < answer < 10) {
+   console.log("Needs to be between 1 and 10");
+ }
+}
```
RDegnen, line 15: this while loop is never going to run because answer already equals 7

RDegnen:
4 of 5 not bad

PR #94, opened 2016-05-18
this is for js-basic diagnostic

tina513:
answers for js-basic

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -6,10 +6,9 @@ let answer = true;
 
 while (answer !== undefined) {
   answer = ask("Pick a number between 1 to 10? ");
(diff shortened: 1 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,15 @@
...
+
+while (answer !== undefined) {
+  answer = ask("Pick a number between 1 to 10? ");
+  if (answer < 1 && answer > 10) {
+   console.log("Your number needs to be any number between 1 to 10!");
+   break;
+ }else{
+  continue;
+ }
+}
```
RDegnen, line 14: true is not going to be < 1 or > 10

RDegnen:
4 of 5 not bad

PR #95, opened 2016-05-18
js-diagnostic basics

jacobhereford:
answers for questions 1-4 and 5

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,12 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+null = null
+undefined = undefined
+number = 0,NaN, Infinity
+boolean = true, false
+string = '', "the empty string."
+
 ## Question 2
 
 ```js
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -54,6 +62,7 @@ weather === "sunny";
 
 What are the values of these expressions?  Explain your answers.
 
+for the first one, weather = "sunny", this is an assignment giving the variable weather the assignment sunny. For weather === "sunny", this is a strict equality and when asked it will come back as either true or false instead of an assignment.
```
RDegnen, line 65: is it true or false?

bin/diagnostic.js
```diff
@@ -1 +1,13 @@
...
+let answer = 'not empty';
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+  if (answer === '6') {
+   console.log("You guessed it!");
+   else = ask("Guess my secret. Hint is 1-10.")
+   break;
+  }
+}
```
RDegnen, line 13: it would be better to add something like 

``` js
  console.log('pick a number between 1 and 10');
}
```

RDegnen:
5 out of 5 awesome!

PR #96, opened 2016-05-18
js-basics-diagnostic Tara Favazza

tfavazza:
question 5 incomplete.

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -12,7 +12,7 @@ while (answer !== 5) {
  if(answer > 10) {
    console.log("your guess has to be a number between 1-10.")
  }
- if (answer === 5) {
+ if (answer == 5) {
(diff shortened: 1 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,25 @@
...
+   console.log("Your guess has to be a number between 1-10.")
+ }
+ if(answer > 10) {
+   console.log("your guess has to be a number between 1-10.")
+ }
+ if (answer === 5) {
+  console.log("You guessed it!");
+  break;
+}
+}
```
RDegnen, line 19: Good, but you could shorten the if statement by doing 

``` js
  console.log("Your guess has to be a number between 1-10.");
}
```

That is saying if the answer is less than 1 OR greater than 10

RDegnen:
5 of 5 nice!

PR #97, opened 2016-05-18
Response for u

computingbets:


README.md
```diff
@@ -53,7 +54,8 @@ weather === "sunny";
 ```
 
-
+1st value is a string assigned to the variable. 2nd is a false boolean value
+because weather strictly does not equal the string sunny.
```
RDegnen, line 58: it will return true because the value of weather is the string 'sunny'.

So when you do `weather ==== 'sunny'` it will return true because weather is 'sunny'

bin/diagnostic.js
```diff
@@ -1 +1,19 @@
...
+  answer = ask("Guess my secret? ");
+  if (answer === '7') {
+   console.log("You guessed it!");
+   break;
+ } else if (answer !== >=1 && <=10){
+   console.log("Pick a number between 1 and 10")
+
+ }
+}
+```'use strict';
```
RDegnen, line 19:
```
'use strict';

var answer = 'not empty';

while (answer !== '') {
  if (answer >=1 && answer <=10) {
   break;
 }else{
   console.log('Please try a number between 1 and 10')
 }
```

RDegnen:
3 of 5

PR #98, opened 2016-05-18
Response

huey1021:


README.md
```diff
@@ -18,6 +18,8 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+string-josh, number-2, boolean- true/false, null, undefined-no value, symbol
+
```
RDegnen, line 22: string would be `"josh"` not `josh`

bin/diagnostic.js
```diff
@@ -1 +1,13 @@
...
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+  if (answer >= 1 && <= 10) {
+   console.log("Your guess needs to be between 1 and 10");
+   break;
+  }
```
RDegnen, line 12: `(answer >= 1 && answer <= 10)` is checking if a number is between 1 and 10.

It is saying 'if answer is greater than or equal to 1 AND answer is less than or equal to 10'

RDegnen:
3 of 5

PR #100, opened 2016-05-18
Jon J Question 1

modularjon:
Answer question 1. Finish at lunch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,12 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+number: 1
+string: 'Jon J'
+booloean: true
+null: null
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -60,6 +65,10 @@ weather === "sunny";
 
 What are the values of these expressions?  Explain your answers.
 
+weather = "sunny" assigns the string "sunny" to the declared variable weather.
+weather === "sunny" is a comparator operator that checks whether weather is
+stricly equal to "sunny", retruning false.
```
RDegnen, line 70: it returns true because the value of weather is "sunny". It is comparing the value of weather, not the name weather

RDegnen:
4 of 5  not bad

PR #107, opened 2016-07-13
Complete JS Basics Diagnostic

gabescarbrough:
I answered the questions to the diagnostic. All but #5 I believe are correct. #5
I know is wrong but I didn't have enough time to figure it out.

Comfort: 4
Clarity: 3

I used Google and my class notes to help with this assessment.

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,23 @@
 'use strict';
+
+//We'll learn about require later in the course
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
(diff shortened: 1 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,23 @@
 'use strict';
+
+//We'll learn about require later in the course
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
```
laurenfazah, line 9: Super close! You'd just need some conditional check for the answer actually being your number of choice (which would need to be defined somewhere).

laurenfazah:
4.5/5, great work!

PR #109, opened 2016-07-13
Add answers to questions 1 - 5.

alex-braun:
Answers to questions 1 - 5.

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,8 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+boolean = true,number = 1string = 'Hi',Null = null,Undefined = y is 7
+
 ## Question 2
 
 ```js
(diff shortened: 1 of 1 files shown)


diagnostic.md
```diff
@@ -9,6 +9,8 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+boolean = true,number = 1string = 'Hi',Null = null,Undefined = y is 7
+
```
laurenfazah, line 13: Confused by your example for `undefined`. Any declared but yet-to-be-defined variable is undefined, so I'm assuming you meant something along those lines? Happy to chat if undefined is unclear at all!

diagnostic.md
```diff
@@ -45,21 +49,26 @@ weather === "sunny";
...
-  if (answer === 'SeCrEt') {
+  if (answer >= 1 && answer <= 10) {
    console.log("You guessed it!");
    break;
-  }
+ }
+ else {
+   console.log("Answer must be between 1 and 10!");
+ }
 }
```
laurenfazah, line 72: Looks very close. I think that now you're able to test this code, you should be able to push it over the finish line. Essentially, though, since `answer = 0` and since 0 is falsey, your `while` loop may never meet the true condition to run.

laurenfazah:
4.5/5, great work Alex!

PR #128, opened 2016-08-24
js-basics diagnostic by VV

viohlet:
Comfort: 4
Clarity: 4

I used Google and my class notes to help with this assessment.

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,23 @@
 'use strict';
+
+```js
+//We'll learn about require later in the course
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '7') {
+  answer = ask("Guess my secret? ");
+  if (answer = [i];
+      (10 >= [i] >= 1) {
+   console.log("You guessed it!");
+   break;
+  }
+}
+```
+
+Change the code from this question so that it tests for a number of your choice
+between 1 and 10 instead of the string `'SeCrEt'`.  If `answer` isn't >= 1 and
+<= 10, use `console.log` to display a message saying that the guess needs to
+be between 1 and 10.  Save your code in the file called `bin/diagnostic.js`.
+You can test the code with `node bin/diagnostic.js` in your Terminal.
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -7,7 +7,11 @@ Assume all code executes in strict mode.
 
 ## Question 1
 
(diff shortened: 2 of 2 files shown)


diagnostic.md
```diff
@@ -7,7 +7,11 @@ Assume all code executes in strict mode.
 
 ## Question 1
 
-Please list the JavaScript primitives and give one example of each.
+undefined: child = undefined
+null: empty object = null
+boolean: var x = false;
```
berziiii, line 12: Correct.

diagnostic.md
```diff
@@ -20,6 +24,7 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
+-The value of c is 7. We add the first values of c+d, which is 7, new value assigned to c.
```
berziiii, line 27: Correct.

diagnostic.md
```diff
@@ -32,8 +37,7 @@ y = 10;
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- solution below -->
-
+- The new value of x=3; new value of y=10. When it comes to numbers, they can leave their last value and convert as the circumstances make them change the values. From right to left, x becomes 3 and only after that, y is 10, not affecting the past value of x.
```
berziiii, line 40: Correct.

diagnostic.md
```diff
@@ -43,8 +47,9 @@ weather = "sunny";
 weather === "sunny";
 ```
 
-
+== means equal value and === means equal value and equal type. Therefore, === is more specific.
+weather = "sunny" is true, but
+weather=== "sunny" is false, they are not the same thing.
```
berziiii, line 52: weather is set to the string of "sunny" to when you check for strict equality, does "sunny" === "sunny"? it would be true.

bin/diagnostic.js
```diff
@@ -1 +1,23 @@
...
+
+```js
+//We'll learn about require later in the course
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '7') {
+  answer = ask("Guess my secret? ");
+  if (answer = [i];
```
berziiii, line 11: Where is [i] defined? You were close but you need to reference an answer value and validate off of that.

berziiii:
3 / 5. Not bad, this will sink in more as we go through this week.

PR #129, opened 2016-08-24
Diagnostic.md commit

rosmith76:
Questions 1-4 answered here

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -8,6 +8,11 @@ Assume all code executes in strict mode.
 ## Question 1
 
 Please list the JavaScript primitives and give one example of each.
+Number -0
+String 'the non-empty string'
+Boolean only produces true or false
+null null
+undefined undefined
 
 ## Question 2
 
@@ -19,7 +24,8 @@ c = c + d;
 ```
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
-
+The final value of c is 7. The final line of the code is assigning a new value
+for c based on two previously assigned varialbles.
 
 ## Question 3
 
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -8,6 +8,11 @@ Assume all code executes in strict mode.
 ## Question 1
 
 Please list the JavaScript primitives and give one example of each.
+Number -0
+String 'the non-empty string'
+Boolean only produces true or false
+null null
+undefined undefined
```
berziiii, line 15: Correct.

diagnostic.md
```diff
@@ -19,7 +24,8 @@ c = c + d;
 ```
 
-
+The final value of c is 7. The final line of the code is assigning a new value
```
berziiii, line 27: Correct.

diagnostic.md
```diff
@@ -33,7 +39,9 @@ y = 10;
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
 <!-- solution below -->
-
+After line 35 each value is 3 since the code reads from right to left
+After line 36 y is 10 and x is 4. Since x wasn't stated with it's new value of 3
+it reverts back to 4.
```
berziiii, line 44: the final values would actually be x = 3 and y = 10. Reason being, you are reassigning x to the y value from directly above it and THEN y is reassigned.

diagnostic.md
```diff
@@ -44,7 +52,9 @@ weather === "sunny";
 ```
 
-
+Line 50 is stating that the vairable weather is "sunny"
+Line 51 weather is strictly equal to "sunny". "sunny" expressed any other
+way will not affect the code. For example SUNNY or Sunny
```
berziiii, line 57: correct.

diagnostic.md
```diff
@@ -55,8 +65,8 @@ const ask = require('../lib/ask.js');
 let answer = 'not empty';
 
 while (answer !== '') {
-  answer = ask("Guess my secret? ");
-  if (answer === 'SeCrEt') {
+  answer = ask("Guess my number? ");
+  for (let i => 1; i <= 10; i++)  {
```
berziiii, line 69: You wouldn't do a for loop here. Reason being, you're checking what the user is entering into the terminal so you would need a if statement to check for some criteria, not a for loop.

berziiii:
3 / 5. Not bad, you were very close on the reassignment one.

PR #133, opened 2016-08-24
Add Answers Update Code

nlkassad:
Comfort: 3
Clarity: 3

Resource list
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures
MDN used to confirm primitive types

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,17 @@
 'use strict';
+
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+if (answer < 0 || answer < 11) {
+  console.log('Booo, needs to be between 1 and 10');
+}
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+  if (answer === '6') {
+   console.log("You guessed it!");
+   break;
+  }
+}
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -7,7 +7,8 @@ Assume all code executes in strict mode.
 
 ## Question 1
 
-Please list the JavaScript primitives and give one example of each.
+Primitive types are Boolean (true, false), Null(null), Undefined(has not been
+assigned, undefined), Number(eg 23.123), String(eg 'Woot!'), and Symbol(used for debugging, don't have a good example).
 
 ## Question 2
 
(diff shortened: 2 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,17 @@
...
+  console.log('Booo, needs to be between 1 and 10');
+}
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+  if (answer === '6') {
+   console.log("You guessed it!");
+   break;
+  }
+}
```
berziiii, line 17: Looks good!

diagnostic.md
```diff
@@ -7,7 +7,8 @@ Assume all code executes in strict mode.
 
 ## Question 1
 
-Please list the JavaScript primitives and give one example of each.
+Primitive types are Boolean (true, false), Null(null), Undefined(has not been
```
berziiii, line 10: Correct.

diagnostic.md
```diff
@@ -18,7 +19,9 @@ c = c + d;
 
 ```
 
-After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
+
+= sets the value of the variable to the content on the right, + does a check for
+the item to the right to be a number then adds it to c. Value should be 12.
```
berziiii, line 24: It would be 7 infact. 

``` js
let c = 5;
let d = 2;

c = c + d; // this would reassign the value of c to be 2 + 5 which is 7
```

diagnostic.md
```diff
@@ -31,9 +34,11 @@ y = 10;
 ```
 
-
 <!-- solution below -->
-
+line 1: x is 4
+line 2: x is 4, y is 3
+line 3: x is y, y is 3 so x is 3
+line 4: x is 3, y is 10
```
berziiii, line 42: Correct.

diagnostic.md
```diff
@@ -44,6 +49,9 @@ weather === "sunny";
 ```
 
+line 1: weather is created, is unassigned because it wasn't assigned a value
```
berziiii, line 52: Correct.

berziiii:
4 / 5. Nice work!

PR #134, opened 2016-08-24
Response

knoddin:


bin/diagnostic.js
```diff
@@ -1 +1,16 @@
...
+const ask = require('../lib/ask.js');
+
+let answer =1;
+
+while (answer !== '') {
+  answer = ask("Pick a number!");
+  if (answer>10) || (answer<1){
+   console.log("Bin needs to be between 1 and 10!");
+   break;
+  }
```
berziiii, line 15: Correct but a user wouldn't get prompted if they are correct. However the code would work as intended. 👍

diagnostic.md
```diff
@@ -7,7 +7,13 @@ Assume all code executes in strict mode.
...
 ## Question 1
 
-Please list the JavaScript primitives and give one example of each.
+Please list the JavaScript primitives and give one example of each
+
+string - text read as text, stored as character values within quotations
+boolean - truthy or falsy values, ex. true or false
+null - no value
+number - numbers! -1 0, 2.7
+undefined - value does not have a defnition
```
berziiii, line 16: Correct.

diagnostic.md
```diff
@@ -20,6 +26,8 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
+C = 7 because 5+2=7 (reading left to right for addition); c+d is assigning the value of 7 to c;
```
berziiii, line 29: Correct.

diagnostic.md
```diff
@@ -32,7 +40,7 @@ y = 10;
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- solution below -->
+x is set as 4 initially, but its value changes to the value of y, which is 3. Y is then changed to the value of 10, but x stays at the value of 3 because it is not reassigned.
```
berziiii, line 43: Correct.

diagnostic.md
```diff
@@ -43,7 +51,7 @@ weather = "sunny";
 weather === "sunny";
 ```
 
+the first weather is being assigned the value of "sunny". The second weather has a strict equality with weather, meaning that whatever the value of weather is, it equals "sunny" and also has the same primitive type (string in this case)
```
berziiii, line 54: Correct.

berziiii:
5 / 5. Nice work!

PR #140, opened 2016-10-17
015/master

raq929:
Updates from node-template and diagnostic-template


RealWeeks:
Nice, simple template update

PR #141, opened 2016-10-19
Create second diagnostic of October 19

akradin:
Answer questions provided

comfort 3
clarity 4

Used class notes

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -79,7 +79,7 @@ while (answer !== '') {
      console.log('Guess again.');
    }
    break;
-  }
+ };
 }
 ```
 
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -73,10 +73,10 @@ while (answer !== '') {
   if (answer === '7') {
    console.log("You guessed it!");
    else if (answer <1 && >10){
-     console.log('The answer needs to be between 1 and 10!')
+     console.log('The answer needs to be between 1 and 10!');
    }
    else{
-     console.log('Guess again.')
+     console.log('Guess again.');
    }
    break;
   }
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,12 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+Number - 23
+String - '23'
+Boolean - true
+null - null
+undefined - undefined
+
 ## Question 2
 
 ```js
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -20,6 +26,10 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
+the value of c is 2
+the operator '+' is the addition operator which adds c and d, so in this scenario 5 and 2
+the operator '=' sets the new value of c
+
```
neugierige, line 32: The value of c is in fact 7. Here's a breakdown line by line:
let c = 5 (c is equal to 5)
let d = 2 (d is equal to 2)
c = c + d (c is now assigned the value of c + d which equals 5 + 2, which equals 7)

diagnostic.md
```diff
@@ -32,7 +42,9 @@ y = 10;
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- solution below -->
+after 'x=y', x = 3
+after 'y=10', y = 10
+The variables themselves don't change but their values do
```
neugierige, line 48: I'll give you this one, but please note that you didn't give the values of x and y for the 1st two lines!

diagnostic.md
```diff
@@ -56,13 +70,29 @@ let answer = 'not empty';
...
+   else{
+     console.log('Guess again.');
+   }
    break;
-  }
+ };
 }
 ```
 
+let newAnswer = 7;
```
neugierige, line 86: I'm not sure which answer you intended to submit, but this one is incorrect because it is missing the while loop. Without the while loop, your code will only run once.

neugierige:
3/5. Good job!

diagnostic.md
```diff
@@ -56,13 +70,29 @@ let answer = 'not empty';
 
 while (answer !== '') {
   answer = ask("Guess my secret? ");
-  if (answer === 'SeCrEt') {
+  if (answer === '7') {
    console.log("You guessed it!");
+   else if (answer <1 && >10){
```
neugierige: You are missing another "answer" in this check. It should be (answer <1 && answer >10)

neugierige, line 75: Sorry, I made a mistake. That if check should be if (answer < 1 || answer > 10). 

(answer <1 && answer >10) will always be false because there doesn't exist a number that is less than 1 AND greater than 10.

PR #142, opened 2016-10-19
Response js basics diagnostic

ioncrash:
Respond to #1-4 in diagnostic.js, respond to #5 in bin/diagnostic.js

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,22 @@
 'use strict';
+
+const ask = require('../lib/ask.js');
+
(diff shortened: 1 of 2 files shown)


neugierige:
5/5. Hi Ben, great job! I'm giving you full credit for question 5, but please note that your code's behavior is not quite what we are looking for. First of all, it's really not necessary to put a while loop inside a while loop. Second, after you console.log "Please try a number between 1 and 10." the prompt for the user to enter a guess does not appear again. I would encourage you to come back to this if you find yourself needing something to work on.

PR #144, opened 2016-10-19
Finish day three second diagnostic

Bowers96:
Answered all the questions given with some minor hiccups on the last
question.

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,13 @@
 'use strict';
+
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== 5) {
+  answer = ask("Guess my number? ");
+  if (answer === 5) {
+   console.log("You guessed it!");
+   break;
+  }
+}
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -7,7 +7,18 @@ Assume all code executes in strict mode.
 
 ## Question 1
 
-Please list the JavaScript primitives and give one example of each.
+String: The String global object is a constructor for strings, or a sequence
+of characters.
+
+Number: The Number JavaScript object is a wrapper object allowing you to work
+with numerical values.
+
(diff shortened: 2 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,13 @@
...
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== 5) {
+  answer = ask("Guess my number? ");
+  if (answer === 5) {
+   console.log("You guessed it!");
+   break;
+  }
```
neugierige, line 12: This is all fine, but your code doesn't fulfill the requirement, which is to also check if the answer is between 1 and 10 (inclusive). Please give it shot and let us know if you need help.

diagnostic.md
```diff
@@ -7,7 +7,18 @@ Assume all code executes in strict mode.
...
+
+Number: The Number JavaScript object is a wrapper object allowing you to work
+with numerical values.
+
+Boolean: The Boolean object is an object wrapper for a boolean value true or
+false
+
+Null: Means null.
+
+Undefined: Means undefined.
```
neugierige, line 21: The question asked to give an EXAMPLE of each. 
String: "hi Brandon"
Number: 1
Boolean: true
Null: null
Undefined: undefined

diagnostic.md
```diff
@@ -32,7 +44,7 @@ y = 10;
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- solution below -->
+x = 3, y = 10, the variables do not change but there value does.
```
neugierige, line 47: The question asked for the values after _each_ line of code.

neugierige:
2/5. Brandon, I can tell that you know this stuff pretty well, but you are not reading the questions. Please be careful of that! I hate to see you missing out on points you could've gotten.

PR #145, opened 2016-10-19
Begin solving problem #5 afternoon diagnostic

kiraamaa:
Complete most of code for problem #5

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,19 +1,20 @@
 'use strict';
 
-## Question 5
 
 
 //We'll learn about require later in the course
 const ask = require('../lib/ask.js');
 
-let answer = 'not empty';
+let answer = NaN;
 
-while (answer !== '') {
-  answer = ask("Guess my secret? ");
-  if (answer >=1 && answer <=10) {
-   console.log("You guessed it!");
-   break;
- }else{
+while (answer !== NaN) {
+  answer = +(ask("Guess the number? "));
+  if (answer > 10 || answer < 1) {
    console.log('Please try a number between 1 and 10');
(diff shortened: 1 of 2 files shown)


kiraamaa:
Comfort: 4
Clarity: 3

I used Google, CodeAcademy, and my class notes to help with this assessment.

bin/diagnostic.js
```diff
@@ -1 +1,20 @@
 'use strict';
+
+
+
+//We'll learn about require later in the course
+const ask = require('../lib/ask.js');
+
+let answer = NaN;
+
+while (answer !== NaN) {
```
neugierige, line 10: Your code works perfectly, but you should know that you should not be checking the opposite of the initial value of "answer" for this while loop. The reason your code is working is actually because of a strange (and frankly, totally illogical) behavior in Javascript whereby NaN === NaN will evaluate to false. (Yea, I know!)

If you had written, "let answer = 'hello'" and "while (answer !== 'hello')", your loop would never run. Because answer is assigned the value "hello", and the while loop checks whether it is NOT "hello" before there is any opportunity for the value of answer to be changed.

diagnostic.md
```diff
@@ -34,6 +46,7 @@ After each line of code executes, what are the values of x and y?  Do the variab
 
 <!-- solution below -->
 
+Value of x is 3, value of y is 10. The variables themselves do not change, although their values are overridden because we are using the 'let' variable type where the value of variables corresponds to the last value assigned.
```
neugierige, line 49: yes, but we were looking for the value of those variable after _each_ line of code

neugierige:
4/5. Great job! Please be sure to read my comment re: question 5.

PR #149, opened 2016-10-19
js-basics Diagnostics Response

kmg1434:
Answered all questions in diagnostics.md
Completed bin/diagnostics.js script
tested

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -2,15 +2,14 @@
 
 const ask = require('../lib/ask.js');
 
-let answer = 'not empty';
+let answer = NaN;
 
-while (answer !== '') {
-  answer = ask("Guess my number: ");
+while (answer !== NaN) {
+  answer = +(ask("Guess my number: "));
   if (answer < 1 || answer > 10) {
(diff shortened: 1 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,15 @@
 'use strict';
+
+const ask = require('../lib/ask.js');
+
+let answer = NaN;
+
+while (answer !== NaN) {
```
neugierige, line 7: Your code works perfectly, but you should know that you should not be checking the opposite of the initial value of "answer" for this while loop. If you had written, "let answer = 'hello'" and "while (answer !== 'hello')", your loop would never run. Because answer is assigned the value "hello", and the while loop checks whether it is NOT "hello" before there is any opportunity for the value of answer to be changed.

The reason your code is working is actually because of a strange (and frankly, totally illogical) behavior in Javascript whereby NaN === NaN will evaluate to false. Try it!

neugierige:
5/5. Great job, and kudos for following the directions! Please read my comment for question 5.

PR #151, opened 2016-10-19
Completed diagnostic.md questions

paulsevere:
Comfort:4
Clarity:4

I used google to make sure that I did not forget any of the primitive types.
I used the node repl to test some code just to be sure of my answers.

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,15 @@
 'use strict';
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+  if (answer < 0 || answer > 10) {
+    console.log('You guess needs to be between 1 and 10')
+  }
+  if (answer === '4') {
+    console.log("You guessed it!");
+    break;
+  }
+}
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,14 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+number: 1
+string: 'this is a string'
+boolean: true
+undefined: var a;
+null: null;
+symbol: Symbol();
+
+
 ## Question 2
 
 ```js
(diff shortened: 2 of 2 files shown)


diagnostic.md
```diff
@@ -9,6 +9,14 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+number: 1
+string: 'this is a string'
+boolean: true
+undefined: var a;
+null: null;
+symbol: Symbol();
```
neugierige, line 17: symbol is not a JavaScript primitive

diagnostic.md
```diff
@@ -9,6 +9,14 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+number: 1
+string: 'this is a string'
+boolean: true
+undefined: var a;
```
neugierige, line 15: Undefined is actually just `undefined`.

diagnostic.md
```diff
@@ -33,7 +43,14 @@ y = 10;
...
 <!-- solution below -->
-
+let x = 4;
+(x == 4, y == undefined)
+let y = 3;
+(x == 4, y == 3)
+x = y;
+(x == 3, y == 3)
+y = 10;
+(x == 3, y == 10)
```
neugierige, line 54: "Do the variables change?  Do their values?"

bin/diagnostic.js
```diff
@@ -1 +1,15 @@
...
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+  if (answer < 0 || answer > 10) {
+    console.log('You guess needs to be between 1 and 10')
+  }
+  if (answer === '4') {
```
neugierige, line 11: Your code works perfectly. However, the solution can be optimized by using just 1 if-statement instead of 2. Multiple if-statements come in handy when you are assessing properties that are independent of each other. Since both of these are testing the value of a, you could just use an "else" clause.

diagnostic.md
```diff
@@ -19,7 +27,9 @@ c = c + d;
 ```
 
-
+c == 7;
```
neugierige, line 30: I assume you mean that the value of c is equal to 7. The expression you wrote here would actually evaluate to `true`. Do you understand why?

bin/diagnostic.js
neugierige: take a look at this one possible solution:

```

let answer = '';

while (answer !== '5') {
 } else if (answer !== '') {
 }
}
```

neugierige:
3/5. Good job! but please remember read the questions carefully and respond to all parts.

PR #152, opened 2016-10-19
Add/modify files for js-basics-diagnostic

kibs86:
Comfort: 3
Clarity: 4

I used my class notes to help with this assessment.

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -2,14 +2,14 @@
 
 const ask = require('../lib/ask.js');
 
-let answer = 'not empty';
+let answer = NaN;
 
-while (answer !== '') {
-  answer = ask("Guess my number? Hint: It's between 1 and 10. ");
+while (answer !== NaN) {
+  answer = +(ask("Guess my number? Hint: It's between 1 and 10. "));
   if (answer < 1 || answer > 10) {
     console.log("Your guess needs to be between 1 and 10.");
   } else if (answer === 3) {
     console.log("You guessed it!");
-    break;
+    answer = NaN;
   }
 }
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,15 @@
 'use strict';
+
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -19,7 +24,7 @@ c = c + d;
 ```
 
-
+The value of c is 7.  The operators in the last line are equals (=) and addition (+).  C is added to D and then C is set to the result of that equation.
```
neugierige, line 27: Correct, but to clarify, a single equal sign is "assignment", meaning the value is c is being assigned the value of `c + d`

bin/diagnostic.js
```diff
@@ -1 +1,15 @@
...
+
+while (answer !== NaN) {
+  answer = +(ask("Guess my number? Hint: It's between 1 and 10. "));
+  if (answer < 1 || answer > 10) {
+    console.log("Your guess needs to be between 1 and 10.");
+  } else if (answer === 3) {
+    console.log("You guessed it!");
+    answer = NaN;
+  }
+}
```
neugierige, line 15: Very close! When I ran your code and entered `3`, it console displayed "You guessed it!" but prompted me again to enter a guess. One way you can fix this is to add a `break;` after line 13. That means that after the code in the `else` clause has run, we will exit out of the loop. Alternatively, you can set your while loops to check is `answer !== 3`.

Also, very importantly, you should not be checking the opposite of the initial value of "answer" for this while loop. If you had written, "let answer = 'hello'" and "while (answer !== 'hello')", your loop would never run. Because answer is assigned the value "hello", and the while loop checks whether it is NOT "hello" before there is any opportunity for the value of answer to be changed.

The reason your code is working is actually because of a strange (and frankly, totally illogical) behavior in Javascript whereby NaN === NaN will evaluate to false. Try it!

neugierige: here's one possible solution, FYI:

```

let answer = '';

while (answer !== '5') {
 } else if (answer !== '') {
 }
}
```

neugierige:
4/5. Great job! Please read over my comments for Q5.

PR #153, opened 2016-10-19
complete diagnostic for javascript basics

sujkid:
The diagnostic for javascript basics submitted

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,8 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+primitives are data types. example of a javascript parameter is boolean.
+
 ## Question 2
 
 ```js
@@ -20,6 +22,12 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
+c = 7;
+Operators are = and +
+operator '=' assigns the value on the right of the operator to the vale on the
+left of the operator. In this case, it evaluates the value of c + d and
+assigns it to c.
+operator '+' performs addition.
 
 ## Question 3
 
@@ -32,8 +40,8 @@ y = 10;
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- solution below -->
-
+x = 3;
+y = 10;
 
 ## Question 4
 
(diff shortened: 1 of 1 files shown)


diagnostic.md
```diff
@@ -9,6 +9,8 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+primitives are data types. example of a javascript parameter is boolean.
+
```
neugierige, line 13: We are looking for a list of _all_ JavaScript primitives, and an example of _each_ one. Take a look at the lesson on JS Basics again: https://github.com/ga-wdi-boston/js-basics

diagnostic.md
```diff
@@ -32,8 +40,8 @@ y = 10;
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- solution below -->
-
+x = 3;
+y = 10;
```
neugierige, line 44: The values of x and y are 3 and 10, respectively, after line 38, but the question asked for the values after _each_ line of code. Also, "Do the variables change?  Do their values?"

diagnostic.md
```diff
@@ -45,22 +53,32 @@ weather === "sunny";
...
+if( answer < 1 && answer > 10 )
+{
+  console.log( "Guess needs to be between 1 and 10" )
+}
+else {
+  while ( answer >= 0 && answer <= 10 )
+  {
+      console.log( "You guessed it!" );
+      break;
+    }
```
neugierige, line 78: Did you have a chance to test your code? A few things: 
1) to implement the behavior we are looking for, the while loop should be outside of the if statement. The reason is that "while" we don't have the input we are looking for, the user should be repeatedly prompted to enter a guess.
2) when you test your code, you'll see that you had a syntactical error with your curly brackets. This is a _very_ common mistake and you'll do this many times over in the next few weeks
3) in line 69, you are check whether `answer` is less than 1 AND greater than 10. Is that actually possible? 
4) `break;` is telling the code to stop running the loop - to "break" out of it. Because you have that directly inside your while loop, when the loop runs for the first time, it would encounter that `break` and immediately exit the loop.

neugierige: here's 1 possible solution, please read this over and make sure you understand what's happening:

```

let answer = '';

while (answer !== '5') {
 } else if (answer !== '') {
 }
}
```

neugierige:
2/5. Hi Suja, I would encourage you to review the lesson on JS Basics, and study the solution for Q5. Let us know if you need help!

PR #154, opened 2016-10-19
Created first JS Code answered diagnostic

jbuzzanco:
Changed the file for diagnostic.js and changed the diagnostic

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,16 @@
 'use strict';
+
+
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+  if (answer === '9') {
+   console.log("You guessed it!");
+   else if (answer <= 1 || answer >= 10);
+   console.log("Guess beween 1 and 10")
+   break;
+  }
+}
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,12 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+1. Number ex- 1
+2. String ex 'bob'
+3. Boolean ex true
+4. Undefined ex undefined
+5. Null ex null
+
 ## Question 2
 
 ```js
@@ -19,6 +25,9 @@ c = c + d;
 ```
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
+
+7
+they are arithmetic operators. = means equals and + means addition
 
 
 ## Question 3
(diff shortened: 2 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,16 @@
...
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+  if (answer === '9') {
+   console.log("You guessed it!");
+   else if (answer <= 1 || answer >= 10);
+   console.log("Guess beween 1 and 10")
+   break;
+  }
+}
```
neugierige, line 16: Very close. Your strategy is right on target, but some of the details are off:
1) Make sure that your brackets match. If you are missing a `{` or a `}`, your code won't run properly. Easiest way to do a check is to run your code.
2) `break;` means to get out of the loop. You have it in the clause that says, "Guess between 1 and 10". So when I guessed 89, I got a message saying "Guess between 1 and 10", and then the program quit. 
3) The question said, "If `answer` isn't >= 1 and <= 10" - so 1 and 10 are _included_ in the values we CAN accept. But what you have on line 12 is `if (answer <= 1 || answer >= 10)`.

diagnostic.md
```diff
@@ -20,6 +26,9 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
+7
+they are arithmetic operators. = means equals and + means addition
+
```
neugierige, line 31: Not quite. A single equal sign is "assignment." It is assigning value to `c`.

diagnostic.md
```diff
@@ -32,7 +41,7 @@ y = 10;
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- solution below -->
+x = 3, y = 10. the variable name stays the same; however, the values change.
```
neugierige, line 44: Fine, but please note that we were looking for values after _each_ line of code

diagnostic.md
```diff
@@ -44,6 +53,9 @@ weather === "sunny";
 ```
 
+weather = "sunny"; // this means that weather variables value is sunny
+weather === "sunny"; // this means that weather variables value and type is sunny
```
neugierige, line 57: This expression will evaluate to `true`. It is actually a boolean expression to test whether `weather` is equal to `"sunny"`

bin/diagnostic.js
neugierige: here's 1 possible solution:

```

let answer = '';

while (answer !== '5') {
 } else if (answer !== '') {
 }
}
```

neugierige:
2/5. Hi Joel, please read over my comments and let me know anything is unclear.

PR #156, opened 2016-10-19
Response

coffeeisamazing:


diagnostic.md
```diff
@@ -8,6 +8,11 @@ Assume all code executes in strict mode.
 ## Question 1
 
 Please list the JavaScript primitives and give one example of each.
+Number:	-0
```
neugierige, line 11: I'll give you this one, but FYI there is no such thing as -0. By definition, 0 is neither positive nor negative. The way we test for positive/negative is to say <0 or >0. And 0 is not <0 nor >0.

neugierige:
5/5. Great job! And kudos for answering all the questions!

PR #157, opened 2016-10-19
Respond to diagnostic questions

joaufi:
Comfort: 5
Clarity: 5

I used Google, previous knowledge, and class notes to finish this assignment.

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,19 @@
 'use strict';
+
+const ask = require('../lib/ask.js');
+
+let answer = 7;
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+
+  if (answer < 1 || answer > 10) {
+    console.log('The guess needs to between 1 and 10!');
+  }
+  console.log(answer);
+
+  if (answer == 8) {
+   console.log("You guessed it!");
+   break;
+  }
+}
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,12 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+Number: 1
+String: 'my string'
+Boolean: true
+Null: null
+Undefined: undefined
+
 ## Question 2
 
 ```js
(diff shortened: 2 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,19 @@
...
+  if (answer < 1 || answer > 10) {
+    console.log('The guess needs to between 1 and 10!');
+  }
+  console.log(answer);
+
+  if (answer == 8) {
+   console.log("You guessed it!");
+   break;
+  }
+}
```
neugierige, line 19: Very close! Your approach is right on target, but some of the details here are off:
1) What happens when the user inputs a single empty space? The while loop condition evaluates to `false`, and the program stops entirely. This is what we call an "edge case" - you want to test your code with all possible cases. 
2) Multiple if-statements are useful when testing for properties that are independent of each other. That's not really the case here, but you are checking the value of `answer` in both cases. You could instead use the `else` clause. What you did not _wrong_, but it did not serve your purpose well.

neugierige: Take a look at this one solution:

```

let answer = '';

while (answer !== '5') {
 } else if (answer !== '') {
 }
}
```

diagnostic.md
```diff
@@ -20,6 +26,11 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
+After the code executes c is equal to 5
+= is an assignment operator; it will assign whatever is on the right hand side
+  to what is on the left hand side (such as a variable)
++ is an addition operator; it will add whatever is on the right side to what
+  is on the left side, sometimes doing type conversions to do so.
```
neugierige, line 33: The value of `c` is actually equal to 7. The reason is that on line 23, `c` is assigned the value of `c + d`, which is 5 + 2. Your explanation is correct. Maybe you made a typo?

diagnostic.md
```diff
@@ -33,7 +44,11 @@ y = 10;
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
 <!-- solution below -->
+The value of x is: 3
+The value of y is: 10
```
neugierige, line 48: Fine, but please note that the questions asks for the values of x and y after _each_ line of code.

neugierige:
3.5/5. Your code for Q5 works for all cases _except_ ' ', hence the half credit.

PR #158, opened 2016-10-19
Complete JS Basics diagnostic

ielole:
Added answers to diagnostic and save some code into bin/diagnostic.js.

"fist to five" 
notes from yesterday and the git repo page

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,15 @@
 'use strict';
+
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (i = 0; i < answer.length; i += 1;) {
+  if answer
+  answer = ask("Guess my secret? ");
+  if (answer === 'SeCrEt') {
+   console.log("You guessed it!");
+   break;
+  }
+}
+```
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -20,6 +25,8 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
+c = 7
+'+' concatenates the variables on either side of it and that result is stored in c.
```
neugierige, line 29: "concatenate" is a term we use to describe the combining of strings or characters. In this instance, we are doing a simple add. What about the `=`? What does that do?

bin/diagnostic.js
```diff
@@ -1 +1,15 @@
...
+
+while (i = 0; i < answer.length; i += 1;) {
+  if answer
+  answer = ask("Guess my secret? ");
+  if (answer === 'SeCrEt') {
+   console.log("You guessed it!");
+   break;
+  }
+}
+```
```
neugierige, line 15: It seems like you started on the wrong foot with this question. Below is one possible solution. Please read through it and make sure you get help if you don't understand it. 

```

let answer = '';

while (answer !== '5') {
 } else if (answer !== '') {
 }
}
```

neugierige:
3/5. Hi Ariel, please look over question 5 again!

PR #160, opened 2016-10-19
Response

rickdean:


bin/diagnostic.js
```diff
@@ -1 +1,23 @@
...
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+  if (answer <1 || >10) {
+   console.log("Needs to be between 1 and 10!");
+   break;
+ }else {
+   console.log("You guessed it!");
+  }
+}
+```
```
neugierige, line 17: Your general strategy is on target. I don't know if you had time to test out your code, but you are missing some important details: 
1) Your while loop tests is the answer is _not_ ' '. What happens if the user input is in fact ' ' (meaning, just an empty space)? The loop would not get run.
2) `break;` is a line of code that tells the computer to exit out of the block of code that you are currently in. Currently, that line occurs when `(answer <1 || >10)` evaluates to true. So, when I guessed 78, the code stopped running. 
3) the question prompt asked you to select a number between 1 and 10, which will be the thing we are testing for. The `else` clause in lines 13-14 is not testing for any condition, which means that I can enter any number between 1 and 10, and it will say "You guessed it!".

neugierige: Here's one possible solution. Take a look at this and see if anything is unclear.

```

let answer = '';

while (answer !== '5') {
 } else if (answer !== '') {
 }
}
```

diagnostic.md
```diff
@@ -8,6 +8,12 @@ Assume all code executes in strict mode.
 ## Question 1
 
 Please list the JavaScript primitives and give one example of each.
+string: "Spider-Man"
+number: "47"
```
neugierige, line 12: a Number type is just a number. The quotes you have around 47 would make it a string.

diagnostic.md
```diff
@@ -20,6 +26,8 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
+After it executes the value of c is 7.
+The operators are c and c+d and they are used to assign a value to c
```
neugierige, line 30: The operators in this expression are `=` and `+`. `=` assigns value to whatever that's on the left of it, and `+` is addition.

diagnostic.md
```diff
@@ -32,8 +40,8 @@ y = 10;
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- solution below -->
-
+x=3
+y=10
```
neugierige, line 44: "Do the variables change?  Do their values?"

diagnostic.md
```diff
@@ -45,6 +53,7 @@ weather === "sunny";
 
 What are the values of these expressions?  Explain your answers.
 
+the values are 'sunny'.  In the first instance the value of 'sunny' is assigned to weather.  In the second weather equals sunny.
```
neugierige, line 56: The 3rd line is a Boolean expression that tests whether `weather` is exactly equal to `"sunny"`, which is true.

neugierige:
0/5. Hi Rick, unfortunately none of your answers were completely correct. Some of them may just be careless mistakes, but we want to make sure that you really understand these concepts, because they are foundational to JavaScript. Please review the JS Basics lesson and, as always, ask for help!
You were well on your way with question 5. Let me know if my comments make sense.

PR #161, opened 2016-10-19
Adds response JS diagnostic

jdaly04:
Adds response to JS diagnostic

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,12 @@
 'use strict';
+const ask = require('../lib/ask.js');
+
+let answer = 'x;
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+  if (x > 1 && < 10) {
+   console.log("Guess again, and choose between 1 and 10");
+   break;
+  }
+}
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,15 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+ Number, String, Boolean, Undefined, Null.
+
+ Examples:
+ Number: 3
+ String: 'hello'
+ Boolean: true
+ Undefined: television
+ null: null
+
 ## Question 2
 
 ```js
@@ -19,6 +28,9 @@ c = c + d;
 ```
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
+
+c = 7
+The operators are saying add variable c to variable, and assign the value of "c" to then be equal to that sum.
 
 
 ## Question 3
@@ -32,7 +44,10 @@ y = 10;
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- solution below -->
+x= 3
+y = 10
+
+The variables don't change, but their values do change.
 
 
 ## Question 4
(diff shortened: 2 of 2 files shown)


diagnostic.md
```diff
@@ -9,6 +9,15 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+ Number, String, Boolean, Undefined, Null.
+
+ Examples:
+ Number: 3
+ String: 'hello'
+ Boolean: true
+ Undefined: television
```
neugierige, line 18: I'll give you this one, but to be clear, `television` would only return `undefined` if it indeed has not been defined in the code. An example of Undefined is just `undefined`.

diagnostic.md
```diff
@@ -20,6 +29,9 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
+c = 7
+The operators are saying add variable c to variable, and assign the value of "c" to then be equal to that sum.
+
```
neugierige, line 34: But what are the operators? And I think you made a typo here? Did you mean "add variable c to variable _d_"?

diagnostic.md
```diff
@@ -42,8 +57,12 @@ let weather;
 weather = "sunny";
 weather === "sunny";
 ```
+What are the values of these expressions?
+"sunny" which is a string. It is words in a quotation.
+
```
neugierige, line 62: The 3rd line is a Boolean expression that checks whether the variable `weather` is exactly equal to the string `"sunny"`. In this case, the value of this expression would be `true`.

bin/diagnostic.js
```diff
@@ -1 +1,12 @@
...
+
+let answer = 'x;
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+  if (x > 1 && < 10) {
+   console.log("Guess again, and choose between 1 and 10");
+   break;
+  }
+}
```
neugierige, line 12: Have you gotten help on this question over the past day or so? There are a couple of things I want to point out:
1) line 4 assign `'x` to answer, but what is `x`? It has not been defined.
2) line 8, binary operators need to have 2 components, so the `< 10` part would need some to precede it. 
3) assuming you wrote the syntax correctly, the expression is checking if x is greater than 1 and less than 10, and if it is, the message says, "Guess again, and choose between 1 and 10." What you are trying to do is check if the input is NOT between 1 and 10.

neugierige: Please study this example solution below and make sure you understand what is happening in this code.

```

let answer = '';

while (answer !== '5') {
 } else if (answer !== '') {
 }
}

```

neugierige:
2/5. Hi Julia, I'm glad you made an effort on question 5, which I know was difficult for many people. Please read over my comments and ask for help if anything is unclear.

diagnostic.md
jdaly04: Totally makes sense. I was basically thinking about it in the sense that I had not yet defined "television", but I did talk to Rachel about this in my 1 on 1 today.

bin/diagnostic.js
jdaly04: Rachel and I went through this one in our 1 on 1 today. I had just started trying to work through it during the diagnostic so definitely knew there were a lot of errors! 

The one thing I'm still a little unclear on though is at the end: 

answer = ask("Guess my secret? ")       <-- What is that doing?  Is it accounting for when the answer is correct? So kind of like "otherwise" if the above stuff is not true, that means they guessed the number.  To me, this wasn't even asked of us in the question within the diagnostic though, right? It seems like the code solution has a lot more in it than what was actually asked in the diagnostic question, or am I misunderstanding something? 

We can chat tomorrow about this too!

PR #162, opened 2016-10-19
JS Basics Diagnostic Answers

beckyd4234:
afternoon diagnostic

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,11 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+number 1
+string 'Becky Davis'
+booleans true
+null null
+undefined undefined
 ## Question 2
 
 ```js
@@ -20,7 +25,7 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
-
+7  operators are = and +
 ## Question 3
 
 ```js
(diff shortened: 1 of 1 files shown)


diagnostic.md
```diff
@@ -20,7 +25,7 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
-
+7  operators are = and +
```
neugierige, line 28: what do the operators do?
`=` assigns values to whatever that's on the left of it, in this case `c`
`+` is simply adding the values of c and d

diagnostic.md
```diff
@@ -30,10 +35,10 @@ x = y;
 y = 10;
 ```
 
-After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
-
-<!-- solution below -->
+After each line of code executes, what are the values of x and y?  Do the variables change?  Do their valuolution below -->
 
+x = 3
+y = 10
```
neugierige, line 42: "Do the variables change?  Do their values?"

The variables do not change, but their values do.

diagnostic.md
```diff
@@ -45,6 +50,8 @@ weather === "sunny";
 
 What are the values of these expressions?  Explain your answers.
 
+true
+false
```
neugierige, line 55: `weather = "sunny";` - assigns the value `"sunny"` to the variable `weather`
`weather === "sunny";` - a Boolean expression that checks if `weather` is exactly equal to `"sunny"`, which would evaluate to `true`

neugierige:
1/5. Hi Becky, I would encourage you to really make an attempt at question 5. I think that the class has seen a lot more JavaScript just in the past day, and that you are now better prepared to tackle this problem. Please be sure to let us know if any of my comments seem unclear.

neugierige:
FYI, here is one possible solution to question 5. Please study this and make sure you understand what's going on here.

```

let answer = '';

while (answer !== '5') {
 } else if (answer !== '') {
 }
}

```

PR #184, opened 2016-11-16
Response

patrickthesailorman:
1 out of 5 generally speaking

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,19 @@
-'use strict';
+//We'll learn about require later in the course
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== 'the guess need to be between 1 & 10') {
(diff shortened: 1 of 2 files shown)


JosephinaTavera:
👏 

Quick notes:

Question 1 
 Type | Examples
---- | ----------------
Number | `-0`, 
String | `''`
Boolean | `true`
null | `null`
undefined | `undefined`

Question 2  They also asked what do the operators in the last line of the program did

= reassigns the value of c
- makes that value the addition of c and d

Question 3 also asked:  Do the variables change? Do their values?

The variables do not change, but their values do.

Question 4   Also looking for the values  at each execution of the line. 

let weather;  // undefined
weather = "sunny"; // "sunny"
weather === "sunny"; // true

Question 5   Here is sample answer

let answer = '';

```
while (answer !== '5') {
 } else if (answer !== '') {
 }
}


```

PR #186, opened 2016-11-19
js-basic-diag > should be everything up-to-date

NewBeg71:
NewBeg71 Week 2 (unsure of original day)

comfort 4.5
completeness 4.75

Comfortable with all answers for 1 to 4; also comfortable with the answer for 5; however, not comfortable with saving one file to another with the terminal - also did not test, which is something else I am not comfortable with as the process which is required with the recent diagnostics


tvcrawley:
:white_check_mark:

PR #194, opened 2017-01-11
Complete JS diagnostic

ddobson:
Add answers for diagnostic.md questions and code to bin/diagnostic.js

Comfort: 5
Clarity: 5

I used Node terminal and yesterday's training repo to complete the diagnostic.

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,13 @@
 'use strict';
+
+const ask = require('../lib/ask.js');
+
+let answer = 4;
+
+while (answer !== 4) {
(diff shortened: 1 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,13 @@
 'use strict';
+
+const ask = require('../lib/ask.js');
+
+let answer = 4;
+
+while (answer !== 4) {
```
bernardlee, line 7: This loop never runs because answer is set to 4 and the loop condition, 4 !== 4, evaluates to false.

bernardlee (commented):
4/5

PR #198, opened 2017-01-11
Mauras response

ladypinetree:


diagnostic.md
```diff
@@ -9,6 +9,12 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+undefined ['']
```
bernardlee, line 12: undefined is just undefined and not an empty string.

diagnostic.md
```diff
@@ -32,7 +40,8 @@ y = 10;
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- solution below -->
+the value of x=y (mutable)
+the value of y=10 (immutable)
```
bernardlee, line 45: | Line | x | y |
| - | - | - |
| let y = 3; | 4 | 3 |
| x = y; | 3 | 3 |
| y = 10; | 3 | 10 |

diagnostic.md
```diff
@@ -41,26 +50,20 @@ After each line of code executes, what are the values of x and y?  Do the variab
...
 weather = "sunny";
 weather === "sunny";
-```
 
-What are the values of these expressions?  Explain your answers.
 
+What are the values of these expressions?  Explain your answers.
+true
+weather="sunny"; assigns the string 'sunny' as the value of weather.
+weather ==="sunny"; asks is the value of weather is the same as the string "sunny"; which is true.
```
bernardlee, line 59: | Expression | Value |
| - | - |
| let weather; | undefined |

bernardlee (commented):
2.5/5

PR #200, opened 2017-01-11
Diagnostics for JS.

farrahj99:
Thank you for this practice!

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,20 @@
 'use strict';
+
+const ask = require('./ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+  if (answer === '>= 1 && <= 10') {
+   console.log("Guess needs to be between 1 and 10");
+   break;
+  }
+}
+```
+
+Change the code from this question so that it tests for a number of your choice
+between 1 and 10 instead of the string `'SeCrEt'`.  If `answer` isn't >= 1 and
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -32,7 +37,10 @@ y = 10;
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- solution below -->
+x's value is 4 and y's value is 3.
+The variables do not change.
+Their values do not change.
+The only thing that changed was y = 10.
```
| - | - | - |
| let y = 3; | 4 | 3 |
| x = y; | 3 | 3 |
| y = 10; | 3 | 10 |

diagnostic.md
```diff
@@ -44,7 +52,8 @@ weather === "sunny";
 ```
 
-
+It is true and the value in this expression is sunny because the variable
+weather is storing the values.
```
bernardlee, line 56: | Expression | Value |
| - | - |
| let weather; | undefined |

bin/diagnostic.js
```diff
@@ -1 +1,20 @@
 'use strict';
+
+const ask = require('./ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+  if (answer === '>= 1 && <= 10') {
```
bernardlee, line 9: This condition causes the loop to terminate when the input is the string ">= 1 && <= 10".  If you want to let the user know that the input is out of range, then the correct condition would have been ```answer < 1 || answer > 10```.

bernardlee (commented):
3/5

PR #201, opened 2017-01-11
Diagnostic Commit One

cshea93:
Diagnostic commit: Complete diagnostic and save code in bin/diagnostic.js

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,86 @@
-'use strict';
+# JS Basics Diagnostic
+
+You should save your answers in this Diagnostic.md file, except for your answer to
+question 5 which should be saved in the file `bin/diagnostic.js`.
+
+Assume all code executes in strict mode.
+
+## Question 1
+
+Please list the JavaScript primitives and give one example of each.
+
+Number(-0), String('name'), boolean(true), null(null), undefined(undefined).
+
+## Question 2
+
+```js
+let c = 5;
+let d = 2;
+c = c + d;
+
+```
+
+After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
+
+c = 7. Operator '=' assigns value. Operator '+', in this case, is addition.
+
+## Question 3
+
+```js
+let x = 4;
+let y = 3;
+x = y;
+y = 10;
+```
+
+After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
+
+y=10
+
+x=3
+
+The variables do not change. The value of y is assigned to x, so x = 3. Value of y becomes 10.
+
+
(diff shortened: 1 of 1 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,86 @@
...
+x = y;
+y = 10;
+```
+
+After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
+
+y=10
+
+x=3
+
```
bernardlee, line 41: | Line | x | y |
| - | - | - |
| let y = 3; | 4 | 3 |
| x = y; | 3 | 3 |
| y = 10; | 3 | 10 |

bin/diagnostic.js
```diff
@@ -1 +1,86 @@
...
+```
+
+What are the values of these expressions?  Explain your answers.
+
+let weather;'s value is undefined becasue we did not assign the variable a value.
+
+ The string "sunny" is assigned to the weather variable.
+
+ weather === "sunny"(string) has a boolean value of true.
+
```
bernardlee, line 60: | Expression | Value |
| - | - |
| let weather; | undefined |

bin/diagnostic.js
```diff
@@ -1 +1,86 @@
...
+  if (answer === 5) {
+   console.log("You guessed it!");
+   else if (answer <= 1 || answer >=10){
+     console.log('Guess needs to be between 1 and 10');
+   }
+   break;
+  }
+}
+```
+
```
bernardlee, line 81: This code is supposed to be in bin/diagnostic.js.

bin/diagnostic.js
```diff
@@ -1 +1,86 @@
...
+//We'll learn about require later in the course
+const ask = require('../lib/ask.js');
+
+let answer = '5';
+
+while (answer ) {
+  answer = ask('Guess a number between 1 and 10');
+  if (answer === 5) {
+   console.log("You guessed it!");
+   else if (answer <= 1 || answer >=10){
```
bernardlee, line 74: This will cause a syntax error since there needs to be a matching closing curly brace before the else-if block.

bin/diagnostic.js
```diff
@@ -1 +1,86 @@
...
+     console.log('Guess needs to be between 1 and 10');
+   }
+   break;
```
bernardlee, line 77: Since nothing stops the flow of execution from reaching ```break;```, this loop will always terminate after one iteration.

bin/diagnostic.js
bernardlee, line 74: The condition ```answer <= 1 || answer >= 10``` disregards 1 and 10 as valid inputs.

bernardlee (commented):
3/5

PR #204, opened 2017-01-11
Day 3 Diagnostic - 2nd diagnostic

zchenbonterra:
-Questions answered for JS

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,12 @@
 'use strict';
+//We'll learn about require later in the course
+const ask = require('../lib/ask.js');
+
+let answer = 1;
+  answer = ask("Guess my number.");
+while (answer !== 1) {
+
+  if (answer<1 && answer>10){
+  console.log("The number is within 1 to 10");
+  }
+}
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,13 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+1. number, 1
+2. string, "name"
+3. undefined, let one; one is not defined since no values are assigned.
+4. NaN, undefine+1
+5. null, empty
+6. boolean, true/false
+
 ## Question 2
 
 ```js
@@ -20,6 +27,7 @@ c = c + d;
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
 
+The value of c is 7. The addition operator, +, adds both side together, and the assignment operator, =, assigns a value to its left operand based on the value of its right operand
 
 ## Question 3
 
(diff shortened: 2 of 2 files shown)


diagnostic.md
```diff
@@ -9,6 +9,13 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+1. number, 1
+2. string, "name"
+3. undefined, let one; one is not defined since no values are assigned.
+4. NaN, undefine+1
```
bernardlee, line 15: NaN is actually of the type Number.

diagnostic.md
```diff
@@ -33,6 +41,7 @@ y = 10;
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
 <!-- solution below -->
+x becomes 3, and y becomes 10. Variables change, but their values which are primitives don't change.
```
| - | - | - |
| let y = 3; | 4 | 3 |
| x = y; | 3 | 3 |
| y = 10; | 3 | 10 |

diagnostic.md
```diff
@@ -44,21 +53,22 @@ weather === "sunny";
 ```
 
-
+"sunny" because it assigns the string "sunny" to te variable weather.
+true because the value that variable weather points to is "sunny" and with the comparison operator, ===, it results in true.
```
bernardlee, line 58: | Expression | Value |
| - | - |
| let weather; | undefined |

bin/diagnostic.js
```diff
@@ -1 +1,12 @@
...
+const ask = require('../lib/ask.js');
+
+let answer = 1;
+  answer = ask("Guess my number.");
+while (answer !== 1) {
+
+  if (answer<1 && answer>10){
+  console.log("The number is within 1 to 10");
+  }
+}
```
bernardlee, line 12: This code results in an infinite loop.

bin/diagnostic.js
```diff
@@ -1 +1,12 @@
 'use strict';
+//We'll learn about require later in the course
+const ask = require('../lib/ask.js');
+
+let answer = 1;
+  answer = ask("Guess my number.");
+while (answer !== 1) {
+
+  if (answer<1 && answer>10){
```
bernardlee, line 9: The condition ```answer < 1 && answer > 10``` will not work since it's impossible for ```answer``` to simultaneously be less than 1 and greater than 10.

bin/diagnostic.js
```diff
@@ -1 +1,12 @@
 'use strict';
+//We'll learn about require later in the course
+const ask = require('../lib/ask.js');
+
+let answer = 1;
+  answer = ask("Guess my number.");
```
bernardlee, line 6: The ```ask()``` prompt needs to be inside of the loop.

bernardlee (commented):
3/5

PR #209, opened 2017-01-11
Response

laurpaik-zz:


diagnostic.md
```diff
@@ -33,7 +41,7 @@ y = 10;
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
 <!-- solution below -->
-
+x becomes 3, and y becomes 10. Both variables change when we tell them to because they're not constants, but they have primitive types and are therefore immutable.
```
| - | - | - |
| let y = 3; | 4 | 3 |
| x = y; | 3 | 3 |
| y = 10; | 3 | 10 |

diagnostic.md
```diff
@@ -44,7 +52,8 @@ weather === "sunny";
 ```
 
-
+the value of `weather = "sunny"` is 'sunny' because we assigned the string 'sunny' to the variable weather.
+the value of weather === "sunny" is true because the operative `===` makes a comparison, and because the computer points the value 'sunny' to the variable weather, the two are equal and the statement is true.
```
bernardlee, line 57: | Expression | Value |
| - | - |
| let weather; | undefined |

bin/diagnostic.js
```diff
@@ -1 +1,29 @@
...
+while (answer !== '') {
+  answer = ask("Guess my favorite number!");
+  if (answer <=1 && answer >= 10){
+    return 'Please guess a number between 1 and 10.';
+  } else if(answer === 2) {
+    return 'You guessed it!';
+  } else {
+    return 'Sorry, try again.';
+  }
+}
```
bernardlee, line 29: This loop only runs once no matter what the input is because ```return``` breaks the loop.

bin/diagnostic.js
```diff
@@ -1 +1,29 @@
...
+// Change the code from this question so that it tests for a number of your choice
+// between 1 and 10 instead of the string `'SeCrEt'`.  If `answer` isn't >= 1 and
+// <= 10, use `console.log` to display a message saying that the guess needs to
+// be between 1 and 10.  Save your code in the file called `bin/diagnostic.js`.
+// You can test the code with `node bin/diagnostic.js` in your Terminal.
+const ask = require('../lib/ask.js');
+let answer = 'not empty';
+while (answer !== '') {
+  answer = ask("Guess my favorite number!");
+  if (answer <=1 && answer >= 10){
```
bernardlee, line 22: This condition is impossible.  The value of answer cannot simultaneously be less than or equal to one and greater than or equal to 10.

bernardlee (commented):
3/5

PR #212, opened 2017-01-11
Response

skylarkJ:


diagnostic.md
```diff
@@ -32,7 +39,10 @@ y = 10;
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- solution below -->
+x = 4;
+y = 10;
+
+Variables got assigned new values.
```
bernardlee, line 45: | Line | x | y |
| - | - | - |
| let y = 3; | 4 | 3 |
| x = y; | 3 | 3 |
| y = 10; | 3 | 10 |

diagnostic.md
```diff
@@ -44,6 +54,9 @@ weather === "sunny";
 ```
 
+weather = "sunny"; - variable weather got stored a value called string "sunny" but it doesn't need to be exactly the strictly "sunny".
+
+weather === "sunny"; variable weather stored stricly the string "sunny" - it must be always written like "sunny"
```
bernardlee, line 59: | Expression | Value |
| - | - |
| let weather; | undefined |

bin/diagnostic.js
```diff
@@ -1 +1,15 @@
 'use strict';
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+
+  for (number = 1; number < 10; number++) {
```
bernardlee, line 9: This will result in an infinite loop.

bin/diagnostic.js
```diff
@@ -1 +1,15 @@
...
+    if (1 <= answer <= 10) {
+      console.log("You are correct.");
+    } else {
+      console.log("Try again, your guess must be between 1 and 10 included.");
+    }
+}
```
bernardlee, line 15: There needs to be an additional closing curly brace to match the while loop's opening curly brace.

bin/diagnostic.js
```diff
@@ -1 +1,15 @@
 'use strict';
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+
+  for (number = 1; number < 10; number++) {
+    if (1 <= answer <= 10) {
```
bernardlee, line 10: JavaScript doesn't allow math-like interval notation to be used.  This would need to be written as ```answer >= 1 && answer <= 10```.

bin/diagnostic.js
bernardlee, line 15: Check the solution branch if you would like to see a solution to this problem.

bernardlee (commented):
3/5

PR #213, opened 2017-01-11
Response

novakw267:
I used mozilla dev network to look up some things

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,15 @@
 'use strict';
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+  if (answer === 1 >= 10) {
+   console.log("You guessed it!");
+   break;
+  }
+  else if (answer !== 1 >= 10){
+    console.log('Guess needs to be between 1 and 10');
+  }
+}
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -44,7 +47,7 @@ weather === "sunny";
 ```
 
-
+The first line weather has been assigned sunny. The second line is asking if they're equal, which currently they're.
```
bernardlee, line 50: | Expression | Value |
| - | - |
| let weather; | undefined |

bin/diagnostic.js
```diff
@@ -1 +1,15 @@
 'use strict';
+const ask = require('../lib/ask.js');
+
+let answer = 'not empty';
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ");
+  if (answer === 1 >= 10) {
```
bernardlee, line 8: This will be interpreted as ```answer === (1 >= 10)```, which evaluates to ```answer === false```, which results in a strict comparison between a string and a boolean.

bin/diagnostic.js
```diff
@@ -1 +1,15 @@
...
+   console.log("You guessed it!");
+   break;
+  }
+  else if (answer !== 1 >= 10){
+    console.log('Guess needs to be between 1 and 10');
+  }
+}
```
bernardlee, line 15: Check the solution branch if you would like to see an answer to this problem.

bernardlee (commented):
3/5

PR #218, opened 2017-01-11
my response

alexkarasik:
Comfort: 3
Clarity: 3

I used Google and my class notes to help with this assessment along with the conosle.

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,80 @@
 'use strict';
+# JS Basics Diagnostic
+
+You should save your answers in this Diagnostic.md file, except for your answer to
+question 5 which should be saved in the file `bin/diagnostic.js`.
+
+Assume all code executes in strict mode.
+
+## Question 1
+
+Please list the JavaScript primitives and give one example of each.
+
+The five primtives are:
+boolean: true,
+numbers: 1,
+strings: "strings",
+undefined: undefined,
+null: null
+
+## Question 2
+
+```js
+let c = 5;
+let d = 2;
+c = c + d;
+
+```
+
+After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d;` and explain what they do.
+
+ The value of c is 7. '+' is the operator in this last line and it adds the previously assigned values of c and d together to give us a new value for c.
+
+## Question 3
+
+```js
+let x = 4;
+let y = 3;
+x = y;
+y = 10;
+```
+
+After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
+
+<!-- solution below -->
+X = 3 and Y = 10. X is given the value 3 because the last assignment given to it is y, which was last given a value of 3. After that, y is given a new value of 10. The variables do not change, only their values.
+
+
+## Question 4
+
+```js
+let weather;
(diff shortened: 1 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,80 @@
...
+   console.log("You guess must be between 1 and 10");
+ }
+}
+```
+
+Change the code from this question so that it tests for a number of your choice
+between 1 and 10 instead of the string `'SeCrEt'`.  If `answer` isn't >= 1 and
+<= 10, use `console.log` to display a message saying that the guess needs to
+be between 1 and 10.  Save your code in the file called `bin/diagnostic.js`.
+You can test the code with `node bin/diagnostic.js` in your Terminal.
```
bernardlee, line 80: This is bin/diagnostic.js.  This is where you should have put in the answer to question 5.

bin/diagnostic.md
```diff
@@ -45,21 +54,21 @@ weather === "sunny";
...
-  if (answer === 'SeCrEt') {
+  if (answer >= 1 && answer =< 10) {
    console.log("You guessed it!");
-   break;
-  }
+ } else {
+   console.log("You guess must be between 1 and 10");
+ }
 }
 ```
```
bernardlee, line 74: diagnostic.md should not be in ./bin.

bin/diagnostic.md
```diff
@@ -33,6 +41,7 @@ y = 10;
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
 <!-- solution below -->
+X = 3 and Y = 10. X is given the value 3 because the last assignment given to it is y, which was last given a value of 3. After that, y is given a new value of 10. The variables do not change, only their values.
```
bernardlee, line 44: | Line | x | y |
| - | - | - |
| let y = 3; | 4 | 3 |
| x = y; | 3 | 3 |
| y = 10; | 3 | 10 |

bin/diagnostic.md
```diff
@@ -45,21 +54,21 @@ weather === "sunny";
 
 What are the values of these expressions?  Explain your answers.
 
-
+First weather is just declared. Then its assigned a value of "sunny". Then it is assigned a strict value of "sunny", "sunny" === "sunny", which is true.
```
bernardlee, line 57: | Expression | Value |
| - | - |
| let weather; | undefined |

bin/diagnostic.md
```diff
@@ -45,21 +54,21 @@ weather === "sunny";
...
 ```js
 
 let answer = 'not empty';
-
 while (answer !== '') {
-  if (answer === 'SeCrEt') {
+  if (answer >= 1 && answer =< 10) {
```
bernardlee, line 67: This will result in a syntax error.  The second part, ```answer =< 10```, should be ```answer <= 10```.

bin/diagnostic.md
```diff
@@ -45,21 +54,21 @@ weather === "sunny";
...
    console.log("You guessed it!");
-   break;
```
bernardlee, line 61: There is no breaking condition so this results in an infinite loop.

bernardlee (commented):
3/5

PR #225, opened 2017-03-23
Diagnostic

cpm5041:


diagnostic.md
```diff
@@ -19,7 +23,8 @@ c = c + d
 ```
 
-
+AFTER the code executes, c = 7. It executes from right to left, so the
+expression would be c = 5 + 2, which equates to 7.
```
jordanallain, line 27: Close enough here. Specifically we wanted you to say that the `+` is adding together 5 and 2, and the `=` is assigning the result to `c`. Sounds like you know that though.

diagnostic.md
```diff
@@ -33,7 +38,11 @@ y = 10
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
 <!-- solution below -->
-
+Line:
+1- x=4
+2- y=3
+3- x=3
+4- y=10
```
jordanallain, line 45: These are right but you didn't answer the final two questions.
- The variables don't change, they remain `x` and `y`.
- The values do change as you assign them.

diagnostic.md
```diff
@@ -44,7 +53,10 @@ weather === "sunny"
 ```
 
-
+(weather = "sunny") = "sunny"
+(weather === "sunny") = true
+The first expression equates to sunny as we are assigning weather a string,
+the second one is true as it is comparing the two values and they are equal.
```
jordanallain, line 59: You got these second two correct but you left out the first one.
`let weather` evaluates to `undefined`

bin/diagnostic.js
```diff
@@ -1 +1,15 @@
...
+while (answer !== '') {
+  answer = ask("Guess my secret? ")
+  if (answer < 1 || answer > 10) {
+   console.log("the guess needs to be between 1 and 10!")
+   break
+  }
+  else {
+    console.log("You guessed it!")
+  }
+}
```
jordanallain, line 15: You were on the right track here Chris. When you include the `break` line in your `if` statement it breaks out of the loop completely. So if someone guesses a number less than 1 or greater than 10 it tells you to guess between 1 and 10 but then breaks out of the loop so you can't keep guessing.

You also haven't given the `else` statement the correct number, so if you guess anything between 1 and 10 it tells you that you are right. You can take a look at the solution branch but also reach out if you want to go over it in more detail.

jordanallain:
3/7

You got caught up on a couple of these but seems like you were on the right track. Let me know if you need any more clarification.

🇯🇴 🇯🇴 🇯🇴

jordanallain:
Also forgot to mention that the instructions wanted a branch  named `response` so pay close attention to the instructions as your branch was named `diagnostic`

PR #227, opened 2017-03-23
Submitting answers for js-diagnostic

ghost:
bin/diagnostic.js and diagnostic.md have been updated.

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,12 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+number: 5
+string: 'string'
+boolean: true
+null: null
+undefined: undefined
+
 ## Question 2
 
 ```js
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -44,7 +54,8 @@ weather === "sunny"
 ```
 
-
+weather = "sunny"  // weather will have a string value of 'sunny'
+weather === "sunny"  // this will compare to see if weather is 'sunny' and will evaluate to true.
```
jordanallain, line 58: You got these two right but you left out the first part.
`let weather` will evaluate to undefined but I could see why you left this line out.

jordanallain:
6/7

Good work here Joe, looks like you ran out of time on writing the function. You don't want to `break` after they give a wrong answer or it will break out of the while loop completely and they can't guess again. Take a look at the solution and if you need any help understanding it let me know.

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #228, opened 2017-03-23
JS Basics Diagnostics

vanduong28:
Response to JS diagnostics with changes in two files.

Comfort: 5
Clarity: 4

I used my class notes and the Google Chrome console feature to assist me on this diagnostic.

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,13 @@
 'use strict'
+const ask = require('../lib/ask.js')
+
+let answer = 'not empty'
+
+while (answer !== '') {
+  answer = ask('Guess my number')
+  if (answer >= 1 && answer <= 10) {
+    console.log('You guessed it!')
+  } else {
+    console.log('Guess needs to be between 1 and 10')
+  }
+}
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -44,6 +56,8 @@ weather === "sunny"
 ```
 
+weather = "sunny" assigns the string 'sunny' to the variable weather
+weather === "sunny" evaluates if the value stored in the weather variable is 'sunny' and will return a boolean (true or false)
```
jordanallain, line 60: Good. Also the first line `let weather` evaluates to `undefined`

bin/diagnostic.js
```diff
@@ -1 +1,13 @@
...
+let answer = 'not empty'
+
+while (answer !== 8) {
+  answer = ask('Guess my number')
+  if (answer >= 1 && answer <= 10) {
+    console.log('You guessed it!')
+  } else {
+    console.log('Guess needs to be between 1 and 10')
+  }
+}
```
jordanallain, line 13: Great job Van! You were so close. Just to pretty it up I would add a `?` and a blank space to the end of the string `'Guess my number'` 

Then you also want to add a `break` statement after printing `'You guessed it!'` otherwise it will continue looping forever!

jordanallain:
6/7

Great work Van, keep it up you're doing awesome.

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #229, opened 2017-03-23
Response - JS Diagnostic

alzateja:
Comfort: 5
Clarity: 4
Here is my completed JS diagnostic.  I tried to answer from memory.
I did compose JS in repl.it compiler. Two commits because I accidentally removed some code.

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,12 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+string = "Boston"
+number = 12
+boolean = true
+undefined = undefined  (like in an undefined variable)
+null = null (like in a defined variable, that is assigned an empty value)
+
 ## Question 2
 
 ```js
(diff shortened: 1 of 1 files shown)


diagnostic.md
```diff
@@ -45,6 +69,14 @@ weather === "sunny"
...
 What do these expressions evaluate to?  Explain your answers.
 
+weather = "sunny"
+- Assigns the value "sunny" to the variable weather
+- No return
+
+weather === "sunny"
+- This statement checks to see if the weather variable is equal to "sunny"
+- Since we assigned "sunny" to weather previously, this evaluates or returns true (boolean)
+
```
jordanallain, line 79: Good, also `let weather` evaluates to undefined.

jordanallain:
5/7

Good job Jayson, looks like maybe you ran out of time. The instructions asked you to write the function in `bin/diagnostic.js` so watch out for that. Also in your function you have the correct answer as the string `'SeCrEt'` but you want it to be a number between 1 and 10. Otherwise good work and let us know if you need extra help with anything.

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #230, opened 2017-03-23
Responses - JS Diagnostic

andyjzhong:


diagnostic.md
```diff
@@ -33,7 +40,7 @@ y = 10
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
 <!-- solution below -->
-
+Once the code is executed, x equals 3 and y equals 10. The variables x and y themselves don't change but their values do.
```
jordanallain, line 43: Right, we also wanted you to list the initial values of `x` and `y` but I think you probably knew what they were.

diagnostic.md
```diff
@@ -44,6 +51,12 @@ weather === "sunny"
 ```
 
+The expressions evaluate to the follow:
+    sunny
+    true
+
+You're creating a variable called weather and setting it equal to the string "sunny".
+Then, you're checking to see whether the variable weather equals "sunny" and indeed it does, therefore it returns "true"
```
jordanallain, line 59: Yes, also the first line `let weather` evaluates to undefined.

jordanallain:
5/7

Great work Andy! Pay close attention to the instructions because you did the work on the `master` branch when you should have made a `response` branch.

Also in your function you have your `break` statement after they give an incorrect answer, which stops the loop so you can't guess again. You want to have it `break` the loop only after they guess the correct answer.

Last your prompt string ends in a `!` when it should end in a `? ` with a space after it so people can type their answer with a gap between the characters. This last one isn't a huge deal but just things to think about so your code is as clear as possible!

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #233, opened 2017-03-23
Response

jim-moody:
5/5.  Felt fine doing these problems although I almost ran out of time. 

I used Google to look up the primitive types because I wasn't sure if there just 5 or not.

Also used the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators#Operator_precedence) to look up if the "=" sign was an operator and found that it was an assignment operator.

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,16 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+String = "Jim"
+
+Boolean = true
+
+Number = 100
+
+undefined = undefined
+
+null = null
+
 ## Question 2
 
 ```js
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -33,7 +47,11 @@ y = 10
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
 <!-- solution below -->
+x = 3
+
+y = 10
 
+The variables don't change but their values do.
```
jordanallain, line 54: Hey Jim we wanted you to note on all four lines what the variables values were but looks like you knew what they were so good job.

jordanallain:
6/7

Looks like you didn't finish the `while` loop for the guessing game. If you need any help with this let us know, you can check out the solution branch if you want. Great job with the questions though!

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #234, opened 2017-03-23
Submission for JS-diagnostic

danielleletarte:
Comfort: 3
Clarity: 3

I used my class notes for this assignment. I'm not clear if the loop was designed to continue asking us to guess if we don't guess correctly. If I guess incorrectly once, I am not prompted to guess again. My logic for values less than 1 and greater than 10 appears to work, though.

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,13 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+String - 'danielle'
+Number - 5
+Boolean - true
+Null - null
+Undefined - undefined
+
+
 ## Question 2
 
 ```js
@@ -19,6 +26,11 @@ c = c + d
 ```
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d` and explain what they do.
+
+c = 7
+Operators
+= - Assigns a value to a variable
++ - Adds two variables c and d
 
 
 ## Question 3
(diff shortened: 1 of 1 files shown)


diagnostic.md
```diff
@@ -45,6 +62,9 @@ weather === "sunny"
 
 What do these expressions evaluate to?  Explain your answers.
 
+let weather --> Declares variable weather
```
jordanallain, line 65: True, this expression technically evaluates to undefined as well.

jordanallain:
5/7

Great job Danielle you nailed all the questions. For the `while` loop, you were supposed to create it in a separate file in the `bin` directory. What you wanted to do for the loop was only put the `break` statement right after they guess the number correctly, that way the loop keeps going until they guess it right.

So instead of:

```
while (answer !== '') {
  answer = ask("Guess my secret? ")
  if (answer === '5') {
   console.log("You guessed it!")
  }
  else if (answer < 1 || answer > 10) {
    console.log("Guess needs to be between 1 and 10")
  }
  break
}
```

it should be

```
while (answer !== '') {
  answer = ask("Guess my secret? ")
  if (answer === '5') {
    console.log("You guessed it!")
    break
  }else if (answer < 1 || answer > 10) {
    console.log("Guess needs to be between 1 and 10")
  }
}
```

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #235, opened 2017-03-23
Diagnostic exercise answers from terry

octater:
here is the updated bin/diagnostic.js and diagnostioc.md files.

Comfort: 4
Clarity: 4

I used my class notes to help with this assessment.

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -8,6 +8,11 @@ Assume all code executes in strict mode.
 ## Question 1
 
 Please list the JavaScript primitives and give one example of each.
+Number:	-0, 156
+String: ' ', 'this is a test'
+Boolean: true, false
+null: null
+undefined: undefined
 
 ## Question 2
 
@@ -20,6 +25,9 @@ c = c + d
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d` and explain what they do.
 
+c is 7.
+'=', this operator asigns the value to the right of it to the variable on the left side.
+'+', this operator adds the values from the left and right together
 
 ## Question 3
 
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -32,7 +40,12 @@ y = 10
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- solution below -->
+x=4,
+x=4, y=3
+x=3, y=3
+x=3, y=10
+
+The variable values change after each line.
```
jordanallain, line 48: Correct, variables remain the same but their values change.

diagnostic.md
```diff
@@ -45,6 +58,9 @@ weather === "sunny"
 
 What do these expressions evaluate to?  Explain your answers.
 
+1. variable weather is set to the sting 'sunny' and returns 'sunny'
+2. the variable weather is compared to the string 'sunny' and returns the boolean true
+
```
jordanallain, line 63: Also `let weather` evaluates to undefined.

jordanallain:
6/7

Good job with the questions Conor. Looks like you were on the right track with the `while` loop.

```
while (answer !== '') {
  answer = ask('Guess my secret?')
  if (answer > 0 && answer < 11) {
    console.log('You guessed it!')
    break
  } else {
    console.log('guess needs to be between 1 and 10')
  }
}
```

Currently this only requires you to guess a number between 1 and 10 to get a right answer. Instead we want to choose a certain number and continue looping until they pick that one number. You did handle the case of the number being less than 1 or greater than 10 though so good job there. If you want to go over this at all let one of us know!

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #236, opened 2017-03-23
Response

khan-ngo:
3/5 

I can find my way around, but I'm not very comfortable...

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,16 +9,24 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+String: let variable = "string"
+number: let variable = 2
+null: let variable = ""
+undefined: let variable
+
+
 ## Question 2
 
 ```js
 let c = 5
 let d = 2
 c = c + d
-
 ```
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d` and explain what they do.
+
+c = 7
+the expression on the right is evaluated before the "=" assignement
 
 
 ## Question 3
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -9,17 +9,25 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+String: let variable = "string"
+number: let variable = 2
+null: let variable = ""
+undefined: let variable
+
+
```
jordanallain, line 17: An empty string in javascript `''` is actually still a string. Null is only `null`. You have to set it to `null`.

So:

```
let jordan = ''
// jordan -> string
let jordan
// jordan -> undefined
let jordan = null
// jordan -> null
```

diagnostic.md
```diff
@@ -9,17 +9,25 @@ Assume all code executes in strict mode.
...
 let d = 2
 c = c + d
-
 ```
 
+c = 7
+the expression on the right is evaluated before the "=" assignement
+
```
jordanallain, line 30: The two operators are `=` and `+`

`=` assigns a value
`+` combines two values

diagnostic.md
```diff
@@ -45,6 +55,7 @@ weather === "sunny"
 
 What do these expressions evaluate to?  Explain your answers.
 
+true ; because weather holds the value "sunny", and truthy and truthy is a true
```
jordanallain, line 58: Okay, we also wanted you to list what the first two lines evaluated to but I'll assume you knew!

jordanallain:
5/7

Great job with the `while` loop! You missed a couple of the instructions in the questions though so read carefully.

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #238, opened 2017-03-23
Response

lisamsmith100:
js Diagnostic responses, diagnostic.md and diagnostic.js

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,9 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+<!-- Q1 ANSWER:  Undefined, string, Boolean, number, NULL -->
+
+
 ## Question 2
 
 ```js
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -9,6 +9,9 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+<!-- Q1 ANSWER:  Undefined, string, Boolean, number, NULL -->
+
+
```
jordanallain, line 14: We also wanted you to list an example for each type.

diagnostic.md
```diff
@@ -32,7 +39,7 @@ y = 10
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- solution below -->
+<!-- Q3 ANSWER:  value of x:  3, value of y:  10-->
```
jordanallain, line 42: Yes and the variables don't change but the values they hold do.

diagnostic.md
```diff
@@ -45,6 +52,8 @@ weather === "sunny"
 
 What do these expressions evaluate to?  Explain your answers.
 
+  //
+<!-- Q4 ANSWER:  true, variable called weather IS EQUAL TO the string "sunny" (this returns true or false)-->
```
jordanallain, line 56: Right and the first line `let weather` evaluates to undefined.

jordanallain:
5/7

Good job Lisa. You were super close with the `while` loop, but you should have put your `break` statement right after they guess the number correctly, that way it keeps looping until they guess right!

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #239, opened 2017-03-23
Answers for diagnostic

rrdaniels85:
I used my class notes for this. 4.5/5. I understood most of it but on the last question for some reason I couldn't get it to run without putting the number I wanted them to guess in quotes. This doesn't make sense given it worked for the numbers in the else statement without the quotes. I am not sure why it required the quotes.

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,15 @@
 'use strict'
+const ask = require('../lib/ask.js')
+
+let answer = 'not empty'
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ")
+  if (answer === '5') {
+   console.log("You guessed it!")
+   break
+ } else if (answer < 1 || answer > 10){
+   console.log("Guess needs to be between 1 and 10")
+   break
+ }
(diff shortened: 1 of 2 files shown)


jordanallain:
3/7

Hey Rachel, it looks like you didn't answer any of the questions in the `diagnostic.md` file, does that sound right to you? I didn't see anything in there. As for the while loop you nearly had it perfect, you just only want to include the `break` statement after a correct answer so that it keeps looping until they get it right.

As for the issue about having the number in quotes, I believe this is just because of the node module that we used to let the command line input interact with the file we wrote our code in. We used a module called `readline-sync` and I believe it only takes strings as input which is why we had to use a string to match the user's input. Let me know if this doesn't make any sense!

🇯🇴 🇯🇴 🇯🇴

rrdaniels85:
Hey Jordan,

?That does not sound rgith. I'm not sure what happened there. I answered all the questions bur when I open it now it shows the same thing you saw. Maybe I didn't commit it correctly? Should I redo and resubmit?

________________________________

PR #240, opened 2017-03-23
Response - Ron Larson

rlarsonil71:
Comfort: 5
Clarity: 4

I used the following to help with this assessment:
- https://github.com/ga-wdi-boston/js . (JavaScript Basics)
- Chrome Inspect (use of console tab to enter/test commands)

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,12 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+Number  5
+String  'Ron'
+Boolean true
+null  null
+undefined undefined
+
 ## Question 2
 
 ```js
@@ -20,6 +26,10 @@ c = c + d
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d` and explain what they do.
 
+c = 7
+
+c = c + d
+7 = 5 + 2  (Value of `c` is added (+) to value of `d` and that value is assigned (=) back into variable `c`)
 
 ## Question 3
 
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -20,6 +26,10 @@ c = c + d
 
 After this code executes, what is the value of c?  Please identify the operators in the last line of this program `c = c + d` and explain what they do.
 
+c = 7
+
+c = c + d
+7 = 5 + 2  (Value of `c` is added (+) to value of `d` and that value is assigned (=) back into variable `c`)
```
jordanallain, line 32: The way you worded this was a bit tough to understand but I think you got it!

diagnostic.md
```diff
@@ -45,6 +57,9 @@ weather === "sunny"
 
 What do these expressions evaluate to?  Explain your answers.
 
+'sunny'   (Assignment operator = is used; String `sunny` is assigned to variable `weather`)
+
+true  (test for equality === is used; String `sunny` is compared to value in variable `weather` which is set to `sunny`.  Both sides of equality operator equal so return is `true`)
```
jordanallain, line 62: Yes, also `let weather` evaluates to undefined.

jordanallain:
6/7

Great job with the questions Ron. You didn't quite finish the `while` loop though. Ideally you would only put the `break` statement in when they get the answer right so that it keeps looping otherwise. You never handled a win condition maybe you ran out of time. If you need any help understanding it let me know!

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #242, opened 2017-03-23
Answered Questions for diagnostic on js

jscohen:


diagnostic.md
```diff
@@ -44,23 +57,32 @@ weather === "sunny"
 ```
 
-
+weather = 'sunny' assigns the string 'sunny' to the variable weather.  It will evaluate to 'sunny' because the variable was instantiated on the previous line.
+weather === 'sunny' will evalue to true.  The === compares weather and 'sunny' and returns true if they are identical, which they are.
```
jordanallain, line 61: Yes and `let weather` evaluates to undefined.

bin/diagnostic.js
```diff
@@ -1 +1,19 @@
...
+// We'll learn about require later in the course
+const ask = require('../lib/ask.js')
+
+let answer
+let solution = 5
+
+while (answer !== 5) {
+  answer = ask("Enter a number between 1 and 10: ")
+  console.log(answer)
+  if (answer == 5) {
```
jordanallain, line 11: Try to always use `===` instead of `==`.

jordanallain:
7/7

Good job Jon. Try to use `===` instead of `==`, I know you used it this time because it wouldn't work with the `===`. If you had changed the answer to the string `'5'` then you could have still used `===` because `readline-sync` takes input as a string which is what you were getting hung up on.

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #244, opened 2017-03-23
Response

angmas:


diagnostic.md
```diff
@@ -8,6 +8,12 @@ Assume all code executes in strict mode.
 ## Question 1
 
 Please list the JavaScript primitives and give one example of each.
+Not a number
+Number
+String
+Boolean
+Null
+Undefined
```
jordanallain, line 16: Believe it or not `NaN` is technically a `number` in javascript as silly as that sounds. So there are only 5 primitives, we also wanted an example of each one.

diagnostic.md
```diff
@@ -33,18 +40,22 @@ y = 10
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
 <!-- solution below -->
-
+x=3 and y=10. The values of x and y change. I'm not sure what the question "do the variables change" means.
```
jordanallain, line 43: The variables remain `x` and `y` but the values they hold change.

jordanallain:
5/7

Good work Angela! You got super close with the `while` loop, let me know if you need any help getting the final solution.

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #245, opened 2017-03-23
Submitting response

sjruby:
Comfort: 3
Claritty: 3

I used google and my class notes, I didn't finish number 5.
The while loop is working as I want it to but the if else
is not working correctly.

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,17 +9,26 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+String = 'text of some kind'
+Numbers = 0,NaN
+boolean = true false
+null = null
+undefined - undefined
+
 ## Question 2
 
 ```js
-let c = 5
-let d = 2
+let c = 5 `creates a variable c with value 5`
+let d = 2 'creates a variable d with value 2'
 c = c + d
 
 ```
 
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -32,8 +41,11 @@ y = 10
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- solution below -->
+After the code runs:
+x === 4
+y === 10
```
jordanallain, line 46: `x` is assigned the value of `y` while `y` is still 3 so the final values are `x = 3` and `y = 10`.

diagnostic.md
```diff
@@ -45,6 +57,8 @@ weather === "sunny"
 
 What do these expressions evaluate to?  Explain your answers.
 
+true b/c === is an "evaluation" to test if the variable weather
+is set to 'sunny'
```
jordanallain, line 61: Yes, also `let weather` evaluates to undefined and `weather = 'sunny'` evaluates to 'sunny'

jordanallain:
4/7

Good work Sam, you missed a couple of the directions but I think you would have got them right otherwise. If you need any help with the `while` loop let me know!

🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #246, opened 2017-03-23
Response

awoodrum87:
For the things covered in this diagnostic I am a 4. I think.

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,12 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+undefined -> let undefinedVariable
+null -> let nullVariable = null
+boolean -> let boolVar = true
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -44,7 +55,9 @@ weather === "sunny"
 ```
 
+weather ="sunny" evaluates to sunny as that is what was assigned to the variable weather.
 
+weather === "sunny" evaluates to true because === is a boolean operator asking is it true that weather is 'sunny'
```
jordanallain, line 60: Yes, also `let weather` evaluates to undefined.

jordanallain:
6/7

Good work Angie. You got really close with the `while` loop, if you need any help getting it to work let me know!

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #251, opened 2017-03-23
Amanda's JS Diagnostic

aeabbott:
I needed more time to fix problem # 4.

I think I would have gotten if I had some more time.

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -6,8 +6,13 @@ question 5 which should be saved in the file `bin/diagnostic.js`.
 Assume all code executes in strict mode.
 
 ## Question 1
-
 Please list the JavaScript primitives and give one example of each.
+
+Number : 1
+String: "Amanda is the best"
+Boolen: true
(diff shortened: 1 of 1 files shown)


diagnostic.md
```diff
@@ -32,7 +41,12 @@ y = 10
 
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
-<!-- solution below -->
+let x = 4 --> 4
+let y = 3 --> 3
+x = y --> 3
+y = 10 --> 10
+
+The variables are change each time you assign them a different value.
```
jordanallain, line 49: Close here, the variables actually remain `x` and `y` but their values do change.

jordanallain:
4/7

Looks like you may have run out of time for the `while` loop, if you want to go over the solution at all let me know.

🇯🇴 🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #252, opened 2017-03-23
JS-Diagnostic Assignment completed.

glasneed:
I ran out of time to finish the script and it has a bug in it. When I guess the correct answer, it still loops even though number between 0 and 10

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -9,6 +9,27 @@ Assume all code executes in strict mode.
 
 Please list the JavaScript primitives and give one example of each.
 
+number
+Example is 5
+
+string:
+Example is "I will love WDI...eventually"
+
+boolean:
+This is either a true or a false.
+Used in loop conditions. Enter a while loop while a condition is still true.
+
+Undefined.
+For example, when you declare a variable and don't assign
+it a value.
+let carModel
+
+Null
+This means nothing. The variable has no value. Again can be used for getting out of a loop condition. Exit loop when
+variable is null (rather than equals 0 which is different)
+
+Used Google for UNDEFINED.
+
 ## Question 2
 
 ```js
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -19,6 +40,9 @@ c = c + d
 ```
 
+c = 5 + 2
+c therefore equals 7
+The right side is evaluated first and then assigned to left side of variable c.
```
jordanallain, line 45: Right, the two operators are `=` for assigning value and `+` to add two values.

diagnostic.md
```diff
@@ -33,7 +57,12 @@ y = 10
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
 <!-- solution below -->
+let x = 4  -- So x equals 4
+let y = 3 -- now y equals 3
+x = y -- now x = 3
+y = 10 -- now y = 10
 
+note: validated answer on node.
```
jordanallain, line 65: Correct, and the variables remain `x` and `y` but their values change.

diagnostic.md
```diff
@@ -45,6 +74,10 @@ weather === "sunny"
 
 What do these expressions evaluate to?  Explain your answers.
 
+the weather variable is assigned the string 'sunny' in line 1.
+Then in line 2,  the value is true because 3 equals means "two strings are strictly equal when they have the same sequence of characters, same length, and same characters in corresponding positions."
+
+Resource: Stack overflow for === and running in node.
```
jordanallain, line 80: `let weather` evaluates to undefined

`weather = 'sunny'` evaluates to 'sunny'

`weather === 'sunny'` evaluates to true

jordanallain:
4/7

Running out of time is okay, let me know if you want to go over any of the solution code.

🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #253, opened 2017-03-23
Response

antleo1995:
This was finished in time but i forgot to push it, and instead worked on a few more commits. You should see a commit from the appropriate time.

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -8,6 +8,12 @@ Assume all code executes in strict mode.
 ## Question 1
 
 Please list the JavaScript primitives and give one example of each.
+<!-- solution below -->
+boolean: true
+string: "a string"
+number: 1
+null: null
+undefined: undefined
 
 ## Question 2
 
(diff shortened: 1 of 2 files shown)


diagnostic.md
```diff
@@ -19,8 +25,11 @@ c = c + d
 ```
 
+<!-- solution below -->
+Value of C is 7
 
-
+The last line is evaluating from right to left so after it adds the numbers it
+reassings the value to c
```
jordanallain, line 32: The two operators are `=` for assigning value and `+` for adding two values together.

diagnostic.md
```diff
@@ -33,7 +42,10 @@ y = 10
 After each line of code executes, what are the values of x and y?  Do the variables change?  Do their values?
 
 <!-- solution below -->
-
+Line 1: value of x is 4 y is undefined at this line
+Line 2: value of x is still 4 and now the value of y is 3
+Line 3: the value of x is set to the valu of y; both variables are now 3
+Line 4: x is still 3 and now y is 10
```
jordanallain, line 48: Correct, the variables remain the same but their values change.

jordanallain:
4/7

Hey Tony thanks for telling me about the two separate commits. Looks like you just ran out of time but you were on the right track. Let me know if you need help with the solution code at all.

🇯🇴 🇯🇴 🇯🇴 🇯🇴

PR #254, opened 2017-03-28
Response which is several days late

donpowers:
I struggled with the whole git thing last week and I failed to submit this at that time.


benjimelito:
Nice going, everything looks good to me in here. Let us know if you have any more trouble with Git for submissions.

PR #259, opened 2017-03-29
All questions answered, code successfully executes

d00medman:
Comfort: 5
Clarity: 5

I used the class lectures as my primary resource.

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,16 @@
 'use strict'
+
+const ask = require('../lib/ask.js')
+
+let answer = 'not empty'
+
+while (answer !== '') {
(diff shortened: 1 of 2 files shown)


benjimelito:
Looking good, 😎  6/7

bin/diagnostic.js
```diff
@@ -1 +1,16 @@
 'use strict'
+
+const ask = require('../lib/ask.js')
+
+let answer = 'not empty'
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ")
```
benjimelito, line 8: Looks good, but you want to give the user a specific number to guess in addition to making sure their answer falls within the range

PR #261, opened 2017-03-29
March 29 2017 Bkim JS-diagnostic 

TronB:
final revisions by Bkim 2:28pm
Comfort: 4
Clarity: 4
Used google as resource

### DIFF
--- a/diagnostic.md
+++ b/diagnostic.md
@@ -8,6 +8,12 @@ Assume all code executes in strict mode.
 ## Question 1
 
 Please list the JavaScript primitives and give one example of each.
+the primitives are:
+-strings | ex: "hi"
+-numbers | ex: 8
+-booleans| ex: true
+-null    | ex: null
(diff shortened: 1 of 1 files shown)


diagnostic.md
```diff
@@ -56,10 +67,13 @@ let answer = 'not empty'
 
 while (answer !== '') {
   answer = ask("Guess my secret? ")
-  if (answer === 'SeCrEt') {
+  if (answer >= 1 && answer <= 10) {
    console.log("You guessed it!")
    break
```
benjimelito, line 72: Don't think you need this here, although it won't hurt things

benjimelito:
Nice going! 6/7

diagnostic.md
benjimelito: Also, you want to give the user a specific number to guess in addition to checking to see if their answer is between 1-10

PR #264, opened 2017-03-29
JS diagnostic.md & Question 5 bin/diagnsotic.js Responses

Pouk91:


bin/diagnostic.js
```diff
@@ -1 +1,13 @@
 'use strict'
+//We'll learn about require later in the course
+const ask = require('../lib/ask.js')
+
+let answer = '>= 1 || <= 10'
```
benjimelito, line 5: This should not be a string, as the Javascript interpreter won't be able to read this as code. Something like `if(answer > 10 || answer < 1){console.log("Please guess a number between 1 and 10!")`

benjimelito:
Good work. Please try to get these in before 9:30am in the future! 6/7

PR #268, opened 2017-03-29
Response

bcem2004:
3
3

I used google and the return search on MDN

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,14 @@
 'use strict'
+const ask = require('../lib/ask.js')
+
+let answer = 'not empty'
+
+while (answer !== '') {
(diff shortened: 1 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,14 @@
 'use strict'
+const ask = require('../lib/ask.js')
+
+let answer = 'not empty'
+
+while (answer !== '') {
```
benjimelito, line 6: Make sure you actually enter a number for the user to guess in addition to checking if their answer was between 1-10.

benjimelito:
Answers looks good and I think your code was close. In the future make sure to start your pull request with enough time to get these in before 9:30!

PR #269, opened 2017-03-29
JS Diagnostic submission

nriser:
Comfort: 4
Clarity: 5

For question 5, I could not resolve an error message I was getting from my linter in diagnostic.js. The error message I got was "Parsing error: Assigning to rvalue (Fatal)".

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,15 @@
 'use strict'
+
+const ask = require('../lib/ask.js')
+
+let answer = 'not empty'
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ")
+  if (answer >= 1 && answer =< 10) {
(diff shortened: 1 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,15 @@
 'use strict'
+
+const ask = require('../lib/ask.js')
+
+let answer = 'not empty'
+
+while (answer !== '') {
+  answer = ask("Guess my secret? ")
```
benjimelito, line 8: Make sure that you are actually providing a number for the user to guess, in addition to making sure their guess is between 1-10

benjimelito:
Make sure that you are using the `ask` variable in your code, I think that may have something to do with your error. Also, please make sure to start your pull request with enough time remaining to get these in by 9:30!

PR #275, opened 2017-03-29
Completed questions and attempted to write code :)

rowlandhill:
Comfort: 4
Clarity: 4

used my notes and google to answer these questions, and was comfortable with the questions.  However, the code was difficult for me and I could not get it to work.

edit: after adding the code, I should probably drop my comfort to 3, since the code didn't work :/

### DIFF
--- a/bin/diagnostic.js
+++ b/bin/diagnostic.js
@@ -1,1 +1,19 @@
 'use strict'
+
+const ask = require('../lib/ask.js')
+
+let answer = 7
+while (answer !== 7) {
+  answer = ask("Guess my number? ")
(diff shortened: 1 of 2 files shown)


bin/diagnostic.js
```diff
@@ -1 +1,19 @@
 'use strict'
+
+const ask = require('../lib/ask.js')
```
benjimelito, line 3: Your code is on the right track, you simply need to include this `ask` variable in your function

benjimelito:
Answers look good and code is almost there, but please make sure to start your pull requests with enough time to get these in before 9:30am!

